PR #4, opened 2020-12-14
Dev context menue

thiloSchlemmer:
Adds a context-menue component


taulinger:
Would it be possible to improve the styling (hover, font-size, pointer) and maybe we should close the menue when an entry has been clicked?

taulinger:
Nice! 
I think we should also close the context menue when a users does a click somewhere else (inside map or outside),

PR #7, opened 2020-12-19
Re-structure project

taulinger:


EnBakir (commented):
No objections. Good job. Right decision to use modularization here.

PR #8, opened 2020-12-22
Dev context menu placement

thiloSchlemmer:


taulinger:
Looks good, redux is used in correct manner.

Two little things:
-  the offset does not seem to have an effect in the upper left area
- i don't think we need js-docs comments on private methods and fields (they are not visible in the rendered docs)

PR #11, opened 2020-12-29
Dev modal window

thiloSchlemmer:
Adding a simple modal-window 

as a starting point implemented with click on the logo with nonsense-content.

### DIFF
--- a/src/modules/modal/components/Modal.js
+++ b/src/modules/modal/components/Modal.js
@@ -1,10 +1,9 @@
 import { html } from 'lit-html';
 import { BaElement } from '../../BaElement';
-import { modalClose } from '../store/modal.action';
+import { closeModal } from '../store/modal.action';
 import css from './modal.css';
 import { $injector } from '../../../injection';
 
-export const NO_CONTENT = '';
 /**
  * @class
  * @author thiloSchlemmer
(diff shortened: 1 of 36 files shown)


src/modules/modal/components/modal.css
```diff
@@ -0,0 +1,60 @@
...
+    padding: 2px 16px;
+    color: var(--color);    
+    display:flex;
+    justify-content: center;
+  }
+  
+  /* Modal Content */
+  .modal {
+    display:none;
+    z-index: 100;
```
thiloSchlemmer, line 47: Instead of manually and static setting:
find max z-index in document and set z-index of modal to maxZIndex++
only onWindowLoad() or on every onStateChanged ?

src/modules/modal/components/Modal.js
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	/**
+	 * Closing the modal window by removing class-name and remove inner content 
+	 * @private
+	 */
+	_closeModal() {
+		this._view.classList.remove('modal--active');	
+		this._clearContent();
+		window.removeEventListener('click', () => this._closeModal());
```
taulinger, line 44: Unnecessary, no former registration of this listener

taulinger (changes requested):
Cool component!

I would suggest following changes/actions:

- extend the tests in `TranslationService.test.js` for the the key "modal_close_button"
- we should use the imperative mood to name an action (closeModal vs. modalClose). Btw, "close" or "open" as names are sufficient here, because all actions of this "class" reference the modal's state
- we should rename the action.type to something like "contentChanged" 
- run lint again after merge with master, which contains extended lint rules

Sorry, yesterday i have overseen the most obvious part:

The component is unnecessary complicated, because it does not use the BaElement Lifecycle.
The whole thing can be done with:

<pre>
import { html } from 'lit-html';
import { BaElement } from '../../BaElement';
import { modalClose } from '../store/modal.action';
import css from './modal.css';
import { $injector } from '../../../injection';

/**
 * @class
 * @author thiloSchlemmer
 */
export class Modal extends BaElement {

	constructor() {
		super();
		const { TranslationService } = $injector.inject('TranslationService');
	}

	/**
	 * @override
	 */
	createView() {
		const { visible, title, content } = this._state;

		if (visible) {
			return html`
        		<style>${css}</style>
				<div class='modal modal--active'>
					<div class='modal-content'>
					<div class='modal-header'>
						<h4  class='modal-title'>${title}</h4>
					</div>
					<div class='modal-body'>${content}</div>
					<div class='modal-footer'>
					<ba-button label='${translate('modal_close_button')}' @click='${modalClose}'></ba-button>
				</div>
			</div>
        </div>`;
		}
		return html``;
	}

	/**
	 * @override
	 * @param {Object} store 
	 */
	extractState(store) {
		const { modal: { title, content } } = store;
		const visible = content && title;
		return { title, content, visible };
	}

	static get tag() {
		return 'ba-modal';
	}
}
</pre>

The action which opens the modal can be called either with a String or another HTML-Template:
(see: [Nest and compose templates](https://lit-html.polymer-project.org/guide/writing-templates#nest-and-compose-templates))

Example:
<pre>
const onClick = () => console.log('click');
const payload = { title:'Info',content: html`
 <p>Lorem ipsum dolor sit amet...</p>
` };
modalOpen(payload);
</pre>

src/modules/modal/components/modal.css
taulinger: In my view a static z-index is enough for the moment, we can implement a more sophisticated mechanism when needed later...

src/modules/modal/components/Modal.js
```diff
@@ -17,99 +16,52 @@ export class Modal extends BaElement {
...
 	/**
 	 * @override
-	 * @param {Object} store 
 	 */
-	extractState(store) {
-		const { modal: payload } = store;
-		return payload;
+	onWindowLoad() {
+		// register callback on ba-button element
+		this._root.getElementById('modalclose').onClick = () => closeModal();
```
taulinger, line 50: unneeded, see line 39

taulinger (commented):
Almost done!
In the meantime, I set  the coverage decrease threshold for failure to 2%

src/modules/modal/components/Modal.js
```diff
@@ -17,99 +16,52 @@ export class Modal extends BaElement {
...
+					<div class='modal-content'>
+						<div class='modal-header'>
+							<h4  class='modal-title'>${title}</h4>
+						</div>
+						<div class='modal-body'>${content}</div>
+						<div class='modal-footer'>
+						<ba-button id='modalclose' label='${translate('modal_close_button')}' @click='${close()}'></ba-button>
+						</div>
+					</div>
+    			</div>`;
```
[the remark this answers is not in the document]
thiloSchlemmer: I tried to refactor the code to completly implement your suggestions, but i dont get it to work properly. Mostly the `@click='${closeModal}` does not work. So i stepped back to the recent intermediate (but suboptimal) solution. In the meantime, i figure out, whats the problem is and solve it.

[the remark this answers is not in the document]
taulinger: Have a look at `dev_modal-window-simplified` -branch.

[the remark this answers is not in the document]
thiloSchlemmer: got it

taulinger:
`import './i18n';` is missing  in the module's index.js

taulinger (approved):
Finally done!!

PR #12, opened 2021-01-11
refactor i18n setup

taulinger:
Each module registers its own i18n provider at the TranslationService


thiloSchlemmer (approved):
Nice refactoring!
In a result the tests for TransformationService.js are coerced to the behavior and the module-related tests are closer to the module-specific tests

PR #24, opened 2021-01-12
ShareService

thiloSchlemmer:
a ShareService to share content with other sources like clipboard etc. 

This pull request is a starting point for future implementations


src/injection/config.js
```diff
@@ -15,7 +16,8 @@ $injector
 	.register('CoordinateService', OlCoordinateService)
 	.registerSingleton('EnvironmentService', new EnvironmentService(window))
 	.registerSingleton('StoreService', new StoreService())
-	.registerSingleton('SearchService', new BvvSearchService());
+	.registerSingleton('SearchService', new BvvSearchService())
```
taulinger, line 19: Semicolon is missing. Running lint before committing would auto-fix that

src/modules/map/components/OlMap.js
```diff
@@ -21,6 +21,8 @@ export class OlMap extends BaElement {
 
 	constructor() {
 		super();
+		const { ShareService } = $injector.inject('ShareService');
```
taulinger, line 24: Our Variable names should start with lower case. We can rename within the destructuring expression with
`const { ShareService : shareService } =  `

taulinger (approved):

taulinger:
could be merged (added two suggestions)

PR #25, opened 2021-01-13
Dev toggle button

taulinger:
Toggle component which uses a slot for customization.
For an example of usage  see: ` ThemeToggle.js`


thiloSchlemmer (approved):
Thanks for this example of  slot-usage.
With this in mind i see again no usage of the slot-concept in the modal-window. Especially with the need  to fill the content of the modal window at runtime, or did i miss something?

PR #30, opened 2021-01-13
refactor to more lit-html usage

thiloSchlemmer:
As a result of  my last component and the learning steps, to master the key concepts, here are a refactoring (trial) of the AutoCompleteSearch-Component to get a better understanding of lit-html usage and the BaElement Lifecycle.

The update of the input.value is currently or still not working.


taulinger (commented):
see comment above

taulinger:
ah, yes I see...

You can call   `selectSuggestion()` within the repeater with the current data: something like `selectSuggestion(this._candidates(i))`

Have a look at line 77 in SidePanel.js

thiloSchlemmer:
Here the next step, based on your (@taulinger ) comments...

i fix (or more add) the update of input-value and extend the usage of lit-html functions (directives)

src/modules/search/components/AutocompleteSearch.js
```diff
@@ -78,24 +59,24 @@ export class AutocompleteSearch extends BaElement {
 			if (e.keyCode == 40) {
 				/*If the arrow DOWN key is pressed,
 				increase the currentFocus variable:*/
-				currentFocus++;
+				this._currentFocus++;
```
thiloSchlemmer, line 62: To get it working properly, i have to lift this variable up to an "internal-state", otherwise i have no access from 
`this._clearCandidates`

src/modules/search/components/AutocompleteSearch.js
```diff
@@ -150,12 +122,40 @@ export class AutocompleteSearch extends BaElement {
...
+		};
+		
 		return html`
 		 <style>${css}</style>
 		 <div class="autocomplete">
-			<input id='autoComplete'/>
+			<input id='autoComplete' .value=${inputText} @input=${onInput}/>
+			${this._candidates ? html`<div id='autocomplete-list' class='autocomplete-items'>${repeat(this._candidates, (candidate) => candidate.id, (candidate, index) => html`
+			<div index=${index} @click=${onClickCandidate} >${unsafeHTML(candidate.labelFormated)}</div>
+		  `)}</div>`: html`` }
```
thiloSchlemmer, line 155: update the code with usage of repeat-directive.

its shorter and slightly more readable then:
`const getAutoCompleteList = () => {
			if (this._candidates) {
				const itemTemplates = [];
				const selectSuggestion = (e) => {
					const selected = this._candidates[e.target.closest('div').getAttribute('index')];
					this._onSelect(selected);
					this._clearCandidates();
				};
				for (let i = 0; i < this._candidates.length; i++) {
					itemTemplates.push(html`<div index=${i} @click=${selectSuggestion} >${unsafeHTML(this._candidates[i].labelFormated)}</div>`);
				}
				return html`<div id='autocomplete-list' class='autocomplete-items'>${itemTemplates}</div>`;
			}
			return html``;
		};

		return html`
		 <style>${css}</style>
		 <div class="autocomplete">
			<input id='autoComplete' .value=${inputText} @input=${onInput}/>
			${getAutoCompleteList()} 
		 </div>
		`;`

src/modules/search/components/AutocompleteSearch.js
```diff
@@ -150,12 +122,40 @@ export class AutocompleteSearch extends BaElement {
...
+			this._clearCandidates();
+		};
+		
 		return html`
 		 <style>${css}</style>
 		 <div class="autocomplete">
-			<input id='autoComplete'/>
+			<input id='autoComplete' .value=${inputText} @input=${onInput}/>
+			${this._candidates ? html`<div id='autocomplete-list' class='autocomplete-items'>${repeat(this._candidates, (candidate) => candidate.id, (candidate, index) => html`
+			<div index=${index} @click=${onClickCandidate} >${unsafeHTML(candidate.labelFormated)}</div>
```
taulinger, line 154: Let's try to call `onClickCandidate`  with model (data) of this item

src/modules/search/components/AutocompleteSearch.js
taulinger: 👍

src/modules/search/components/AutocompleteSearch.js
thiloSchlemmer: OK...done.

taulinger (approved):
Nice 😎

Instead of an empty  `html` you can use also `nothing` if you want:
https://lit-html.polymer-project.org/guide/writing-templates#rendering-nothing

PR #35, opened 2021-01-15
Extend render callbacks

taulinger:


thiloSchlemmer (approved):
I guess, i got it :) 

With this change, the ThemeToggle-Button gets working in the showcase, because the callback is registered after the lit-Elements are fully rendered, right?

PR #37, opened 2021-01-15
DEV info button with popup

EnBakir:
A little dummy button with click action, opening a simple popup.


taulinger (changes requested):
Overall, this points strongly into the right direction 👏

Some suggestions from my side:

`Popup.js`
- expose show/hide functionality of this component via property and attribute (see: `Button.js`) or by a method 
- possibly make the position confugurable by an attribute (see: `Button.js`)
- the user specific content should be delivered by a slot (see: `Toggle.js`)

`InfoButton.js`
- adjust the svg -icon (https://fontawesome.com/icons?d=gallery&m=free)

### Testing
- test are missing!!

EnBakir:
> Overall, this points strongly into the right direction clap
> 
> Some suggestions from my side:
> 
> `Popup.js`
> 
>     * expose show/hide functionality of this component via property and attribute (see: `Button.js`) or by a method
> 
>     * possibly make the position configurable by an attribute (see: `Button.js`)
> 
>     * the user specific content should be delivered by a slot (see: `Toggle.js`)
> 
> 
> `InfoButton.js`
> 
>     * adjust the svg -icon (https://fontawesome.com/icons?d=gallery&m=free)
> 
> 
> ### Testing
> 
>     * test are missing!!

Thanks for reviewing this first try of coding! The points you mentioned are very helpful, these are just the things i missed. I will also add tests for this class. :+1:

src/modules/commons/components/popup/Popup.js
```diff
@@ -0,0 +1,110 @@
...
+	 *@private 
+	 *@param {string} right
+	 *@param {string} top   
+	 */
+	_setPosition(right, top) {
+		var popup = this.shadowRoot.getElementById('this-popup');	
+		popup.style.right = right + 'px';
+		popup.style.top = top + 'px';
+		this.render();
+	}
```
taulinger, line 106: Not needed and breaking a clean approach:
All style settings should be done within `createView` dependent from current state e.g. (top, right) of the component

taulinger (commented):
Made some nice progress 😃!

- Have a look at my comment on `Popups.js`
- Tests for `InfoButton.js` are missing

src/modules/commons/components/popup/Popup.js
EnBakir: Thanks for that ! I'll fix this :smile:

src/modules/map/components/infoButton/InfoButton.js
```diff
@@ -0,0 +1,68 @@
...
+		const openPopup = () => {
+
+			// set created popup visible
+			var popup = this.shadowRoot.getElementById('info-popup');
+			if (popup.getAttribute('type') === 'hide') {
+				popup.setAttribute('type', 'show');
+			}
+			else{
+				popup.setAttribute('type', 'hide');
+			}
```
taulinger, line 41: It would be nicer, if we could call methods of Popus.js, like  `open()`.
If you want to test such calls, you have to register the ba-popup -element within the `InfoButton.test.js`

taulinger (changes requested):
Almost there! Tests are good!

Currently, when I want to handle the popup by properties (not by attribute), I have to write
something like
`popup.type = "show"`
I think, it would be easier to use  your component, if the ba-popup element  would also have had methods
like `open()` `close()` `isOpen()`.

src/modules/map/components/infoButton/InfoButton.js
EnBakir: Yeah, maybe that's better that way ! I'll do that :smiley:

src/modules/map/components/infoButton/InfoButton.js
```diff
@@ -29,18 +29,25 @@ export class InfoButton extends BaElement {
...
 			var popup = this.shadowRoot.getElementById('info-popup');
-			if (popup.getAttribute('type') === 'hide') {
-				popup.setAttribute('type', 'show');
+			if (!popup.isOpen()) {
+				popup.openPopup();
 			}
 			else{
-				popup.setAttribute('type', 'hide');
+				popup.closePopup();
 			}
```
taulinger: Is this clause really needed? If true, we also need a test for this clause and we should rename the function ti something
like `togglePopup`

src/modules/map/components/infoButton/InfoButton.js
```diff
@@ -29,18 +29,25 @@ export class InfoButton extends BaElement {
...
 		};
 
+		// close popup on map click
+		window.addEventListener('map_clicked', () => {
+			var popup = this.shadowRoot.getElementById('info-popup');
+			if (popup.isOpen()) {
+				popup.closePopup();
+			}
+		}); 
+
```
taulinger, line 50: We should remove that, because it uses a CustomEvent, which is only fired for demonstration purposes.
A better approach in the future  will be using the state store, e.g. listening to changes of the map position.
Btw: we would also need a test for this section

taulinger (changes requested):
Two more little things...

src/modules/map/components/infoButton/InfoButton.js
EnBakir: That's my fault. This isn't needed. I'll remove that.

src/modules/map/components/infoButton/InfoButton.js
EnBakir: Okey didin't know that. So we can do that after we implement the state store listening to map events.

test/modules/map/components/infoButton/InfoButton.test.js
```diff
@@ -77,4 +80,48 @@ describe('InfoButton', () => {
...
+			element.shadowRoot.getElementById('info-popup').openPopup();
+
+			expect(element.shadowRoot.getElementById('info-popup').getAttribute('type')).toBe('show');
+			expect(element.shadowRoot.getElementById('info-popup').isOpen()).toBeTrue();
+
+			expect(store.getState().map.zoom).toBe(5); 
+
+			changeZoom(4);
+
+			expect(store.getState().map.zoom).toBe(4);
```
taulinger, line 99: Uneeded, we don't want to test the store (also in line 121)

taulinger (changes requested):
Almost done (really 😉)

src/modules/map/components/infoButton/InfoButton.js
```diff
@@ -68,6 +57,18 @@ export class InfoButton extends BaElement {
...
+	extractState(store) {
+		const { map: { zoom, position } } = store;
+		return { zoom, position };
+	}
+
+	onStateChanged() {
+		var popup = this.shadowRoot.getElementById('info-popup');
+		if (popup.isOpen()) {
+			popup.closePopup();
+		}
```
taulinger, line 69: We can omit the if-clause here (also near line 35)
(have also a at your code coverage report for this file: coverage/modules/map/components/infoButton/index.html)

taulinger (approved):
Time to merge 😌

PR #42, opened 2021-01-20
IconButton

thiloSchlemmer:
A first starting point for a reusable IconButton. Not all features we wish to implement, are currently working, so this draft should evolve with the time.

The current approach is implemented with the example-use in the "showcase"-logic (Header.js).


EnBakir (commented):
Good start for this component 👍🏻. Maybe you could add a test case for the slot (look at Toggle.test and on branch dev_infoButton Popup.test)

thiloSchlemmer:
> Good start for this component 👍🏻. Maybe you could add a test case for the slot (look at Toggle.test and on branch dev_infoButton Popup.test)
thx good hint

EnBakir (approved):

taulinger (commented):
At a later moment we should rewrite this component in order to deliver just the svg icon from outside  and not a whole bunch of css classes. 
I don't think we can use the slot approach here.
Maybe we extend that component then and show an icon and text ... but let's wait for specific use cases...

taulinger:
@thiloSchlemmer I would like to suggest closing this PR. If we need it later, we can reopen it ...

EnBakir:
> @thiloSchlemmer I would like to suggest closing this PR. If we need it later, we can reopen it ...

I agree with that.

PR #43, opened 2021-01-20
Dev layer manager

taulinger:
Added the state stuff for layer management. Have look at `/modules/map/store/layers/layers.action.js`


taulinger (commented):
Looks good for now 👍
 Don't forget tests for the i18n provider

src/modules/map/i18n/index.js
```diff
@@ -1,5 +1,7 @@
 
 import { provide } from './map.provider';
+import { layerManagerProvide } from './layerManager.provider';
```
thiloSchlemmer, line 3: The naming here was a starting point from me, but i would prefer a balanced naming for both provide-methods, like  
`mapProvide` and `layerManagerProvide`. 
With the growing of components in the modules with they seperated i18n-providers, there should be a common convention how to deal with it.

test/modules/map/i18n/layerManager.provider.test.js
```diff
@@ -0,0 +1,28 @@
...
+		expect(map.layer_manager_change_visibility).toBe('toggle visibility');
+	});
+
+	it('provides an empty map for a unknown lang', () => {
+
+		const map = layerManagerProvide('unknown');
+
+		expect(map).toEqual({});
+	});
+});
```
thiloSchlemmer, line 28: Here the missing tests for i18n

src/modules/map/i18n/index.js
taulinger: thx

taulinger:
With the latest commits it is possible to add/remove/modify ol-layers on the the map by calling the corresponding action creator from `layers.action.js`.

 The ol-layers are build on data provided by a GeoResourceService, in our case `BvvGeoResourceService,js`. Currently three WmsGeoResources are available: 'bodendenkmal', 'baudenkmal', 'dop80'

Example: To add the 'dop80' georesource, you would call:
 `addLayer('bodendenkmal')`

taulinger:
Synchronization problems are fixed, multiple calls of `addLayer` are possible now.

taulinger:
sorry for the confusion (DPR->PR->DPR)

taulinger:
I think we should  move the `layerItem` folder into the `layerManger` folder, because it is a nested child component, which should only be used in there. What do you think?

thiloSchlemmer:
> I think we should move the `layerItem` folder into the `layerManger` folder, because it is a nested child component, which should be used in there only. What do you think?
> 
> Btw: time well spent on that refactoring 👍


Yes, thats OK. As long as LayerItem is only referenced in Layermanager and is not used otherwise. This would be a good and practical update to our packaging-convention.

thiloSchlemmer:
Problems with LayerManager/LayerItem are fixed.

src/modules/map/components/layerManager/LayerManager.js
```diff
@@ -0,0 +1,156 @@
...
+				}			
+			}			
+		};
+
+		return html`
+			<style>${css}</style>
+			<div class="layermanager overflow-container">
+				<div class='title'>${translate('layer_manager_title')} (${layerCount})</div> 
+				<ul class='layers'>
+                    ${repeat(this._draggableItems, (layerItem) => layerItem.listIndex, (layerItem, index) => html`
```
taulinger, line 128: Here we have still a problem: the id for the item is not unique, that can can cause some confusion during the rendering phase. (from the docu: _keyFunction is a function that takes a single item as an argument and returns a guaranteed unique key for that item._)
Due to this,  you have to call `render()` manually in `LayerItem.js` line 121, which wouldn't be necessary otherwise. 
But that's just the effect, the problem is, that the repeater cannot reuse your items correctly

thiloSchlemmer: Ah...ok...i fix this. I did not realize this as update-render-mechanism, only as an index for a sortable

taulinger: Yes, based on a stable and unique item id, the reapter "knows" which which part of its items have to be re-rerndered.

thiloSchlemmer (approved):

PR #45, opened 2021-01-21
Toggle element: make it possible to register the callback within the markup

taulinger:
It's now possible to register the callback by an event listener within the markup:
`<ba-toggle  @toggle=${onToggle}><span>Toggle me!</span></ba-toggle>`

Have also a look at our showroom...


thiloSchlemmer (approved):
Nice blueprint for a custom-event!

thiloSchlemmer (approved):

PR #50, opened 2021-01-22
update contextmenue with more litify usage

thiloSchlemmer:
Coming back to the start of my lit-html learning-curve :) 
...and refactor this component for a cleaner lit-html approach.


taulinger (approved):
Good refactoring!

PR #56, opened 2021-01-25
add tests for missing test cases (throw error)

thiloSchlemmer:


test/injection/core/unregisteredInjection.test.js
```diff
@@ -0,0 +1,30 @@
+import { $injector } from './testsConfig';
+var mockService =  { get: 'I\'m a mock. ' };
```
taulinger, line 2: 🧐

taulinger (approved):
well done!

test/injection/core/registerTwiceError.test.js
```diff
@@ -0,0 +1,18 @@
+import { $injector } from './testsConfig';
+var mockService =  { get: 'I\'m a mock. ' };
```
taulinger, line 2: 🧐 we don't want vars anymore 😉

thiloSchlemmer: ah yes we make `let`/`const` not `var`... indeed :)

PR #61, opened 2021-02-01
Clickable Icon

taulinger:
Clickable icon component, The svg graphic can be configured via the `icon` attribute / property.
This is not a replacement for the icon-button, which we will need at a later moment.


EnBakir (approved):
Nice icon button :+1:

thiloSchlemmer (approved):

PR #63, opened 2021-02-02
update css and remove unnecessary showcase-code

thiloSchlemmer:
update max-height value for modal-component, to prevent unreachable content

remove unnecessary 'Lorem ipsum' content from showCase


taulinger (approved):
👍

PR #71, opened 2021-02-05
Dev: button zoom to extent

EnBakir:
Button that allows a zoom to extent.


taulinger:
Store/Map infrastructure is set. Next we have to do:
- (more) tests for `ExtentButton.js`
- beautify the layout of the button
- get the default extent from a central service (me)

EnBakir:
Great, then i can continue :+1:

taulinger:
MapService is there...

To get the default extent, you have to `inject('MapService')`, then call `mapService.getDefaultMapExtent()`

src/modules/map/i18n/map.provider.js
```diff
@@ -9,7 +9,8 @@ export const provide = (lang) => {
 				map_info_button: 'Information',
 				map_info_button_help: 'Help',
 				map_info_button_contact: 'Contact',
-				map_info_button_about: 'About us'
+				map_info_button_about: 'About us',
+				map_zoom_extent_button: 'Show whole map'
```
thiloSchlemmer, line 13: I would prefer: 'Zoom to full extent'

src/modules/map/i18n/map.provider.js
```diff
@@ -20,7 +21,8 @@ export const provide = (lang) => {
 				map_info_button: 'Information',
 				map_info_button_help: 'Hilfe',
 				map_info_button_contact: 'Kontakt',
-				map_info_button_about: 'Impressum'
+				map_info_button_about: 'Impressum',
+				map_zoom_extent_button: 'Ganz Bayern anzeigen'
```
thiloSchlemmer, line 25: we can avoid a country-specific text?

thiloSchlemmer (changes requested):
Some hints/questions inline

src/modules/map/i18n/map.provider.js
taulinger: I think, it's OK when we use BayernAtlas specific translations in German

src/modules/map/i18n/map.provider.js
EnBakir: Thanks, that's a good idea !

taulinger: @EnBakir Can you change that asap, please? We need the infrastructure code...😉

taulinger (approved):
Go for a merge

thiloSchlemmer (approved):

PR #73, opened 2021-02-08
Fix info button

EnBakir:
The info icon seemed to be too small for the button, so that there was a white outline visible (thx to @taulinger for reporting). Was fixed by making the icon bigger with the viewBox attribute of the svg-tag. (Funnily width and height (of svg-tag) attribute had no impact on that icon.)


taulinger (approved):
Good to know 👍

thiloSchlemmer (approved):

PR #85, opened 2021-02-16
Select coordinate system 

EnBakir:
Added a select for coordinate system selection.


taulinger:
Good start!

Have a look at #87. Those changes will help you to:

1. Create a separate component just for displaying the coordinates. The MapInfo component is just a (temporary?) container
2. Do not use SRID/EPSG definitions yourself. Get a list of valid codes from the map service by (`#getSridDefinitionsForView`) and let the view iterate over that

taulinger (changes requested):
Good start!

Have a look at #87. Those changes will help you to:

1. Create a separate component just for displaying the coordinates. The MapInfo component is just a (temporary?) container
2. Do not use SRID/EPSG definitions yourself. Get a list of valid codes from the map service by (`#getSridDefinitionsForView`) and let the view iterate over that

src/modules/footer/components/coordinateSelect/CoordinateSelect.js
```diff
@@ -0,0 +1,92 @@
...
+	/**
+	 * @override
+	 */
+	onWindowLoad() {
+		// without this if clause,  this._view equals null
+		if (!this._environmentService.isTouch()) {
+			this._view = this.shadowRoot.querySelector('.select-coordinate');
+			this._selected = this._view.value;
+		} 
+	}
```
taulinger, line 41: Avoid holding HTML elements as local state, just data.
In this case I would suggest two fields which are initially  set within the constructor
-  `_items` (loaded by the MapService )
- `_selectedCode`  the code value of the first item

The whole creation of the View just depends on that two fields. When the select element changes, `_selectedCode` is being updated and the whole view must be re-rendered. The re-rendering has to be done manually

taulinger (commented):
Good overall progress 😀

src/modules/footer/components/coordinateSelect/coordinateSelect.css
```diff
@@ -0,0 +1,39 @@
...
+
+
+.coordinate-label {
+    padding: 10px;
+    /* flex-grow: 100; */
+    display: block;
+}
+
+.select-coordinate {
+    background-color:  var(--primary-color);
```
taulinger, line 22:
```
background-color:  var(--secondary-bg-color);
color: var(--color);
```

test/modules/footer/components/coordinateSelect/CoordinateSelect.test.js
```diff
@@ -0,0 +1,107 @@
...
+				pointerPosition: [1288239.2412306187, 6130212.561641981]
+			}
+		};
+
+		TestUtils.setupStoreAndDi(state, { position: positionReducer });
+		$injector
+			.registerSingleton('TranslationService', { translate: (key) => key });
+		$injector
+			.register('CoordinateService', OlCoordinateService);
+		$injector
```
taulinger, line 28: Let's use a mocked CooridnateService (see: `MapInfo.js`)

taulinger (commented):
Good job! 

Btw, we should remove the zoom level label from the `MapInfo` element...

taulinger:
👍 
Currently the translation service is unused in `CoordinateSelect`. Did you plan to add a title attribute on the select (which would be nice)?

taulinger (approved):

PR #86, opened 2021-02-16
Display map context information

taulinger:
Context Menu which is is opening on a  `contextmenu` event.
Displays coordinates in various projections based on the location of the event.
Currently no support for touch devices (long-press, long-hold)


thiloSchlemmer (approved):
Nice and Elegant structural enhancement for the event-handling!

PR #95, opened 2021-02-20
Fix wrong positioning on first click events

### LINKED ISSUE #94: Contextmenu has wrong position
... on first and second click  in right and/or lower sector of the map

taulinger:
Re-implementation of the positioning mechanism of the `MapContextMenu`. Fixes #94 .

### DIFF
--- a/src/modules/map/components/contextMenu/MapContextMenu.js
+++ b/src/modules/map/components/contextMenu/MapContextMenu.js
@@ -4,7 +4,6 @@ import { BaElement } from '../../../BaElement';
 import css from './mapContextMenu.css';
 import { $injector } from '../../../../injection';
 import { close as closeContextMenu } from '../../store/mapContextMenu.action';
-
 import closeIcon from './assets/x-square.svg';
 
 /**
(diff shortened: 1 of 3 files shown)


src/modules/map/components/contextMenu/MapContextMenu.js
```diff
@@ -22,77 +21,66 @@ export class MapContextMenu extends BaElement {
...
-		const windowHeight = window.innerHeight - (window.innerHeight * offsetBorderInPercent);
-
-		if ((windowWidth - coordinate[0]) < menuWidth) {
-			values.vAlignment = 'right';
-			values.left = coordinate[0] - menuWidth + 'px';
-		}
-		else {
-			values.vAlignment = 'left';
-			values.left = coordinate[0] + 'px';
+		//windiw sector the click event occurred:
```
thiloSchlemmer, line 29: Typo

src/modules/map/components/contextMenu/MapContextMenu.js
```diff
@@ -22,77 +21,66 @@ export class MapContextMenu extends BaElement {
...
+			return 1;
 		}
-		else {
-			values.hAlignment = 'top';
-			values.top = coordinate[1] + vOffset + 'px';
+		else if (coordinate[0] > widthBorder && coordinate[1] > heightBorder) {
+			return 2;
 		}
-		return values;
+		return 3;
```
thiloSchlemmer, line 42: Do we discussed  already named_values for magic numbers like 0-3? For instance `const VIEW_QUADRANT_UPPER_LEFT =0` etc.

thiloSchlemmer (approved):

src/modules/map/components/contextMenu/MapContextMenu.js
taulinger: thx

src/modules/map/components/contextMenu/MapContextMenu.js
taulinger: Good point, but let's wait until we can use static fields. We want to use such named values also outside this single method. From my point of view it's confusing, if we would use member variables for that case, they should represent state...

PR #101, opened 2021-02-23
add a switch for device related snapTolerance

thiloSchlemmer:
adding a private getter-method for snaptolerance depending on
environment information (isTouch) to deliver a better draw-behavior in
small areas on Desktop-Environments on the one hand and
better first-point or last-point snaps on Touch-Environments

### DIFF
--- a/test/modules/map/components/olMap/handler/measure/OlMeasurementHandler.test.js
+++ b/test/modules/map/components/olMap/handler/measure/OlMeasurementHandler.test.js
@@ -106,52 +106,26 @@ describe('OlMeasurementHandler', () => {
 		
 	});
 
-	describe('when activated over olMap', () => {
-		
-		const initialCenter = fromLonLat([11.57245, 48.14021]);
-
-		const setupMap =  () => {
-			return new Map({
-				layers: [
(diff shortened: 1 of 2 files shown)


test/modules/map/components/olMap/handler/measure/OlMeasurementHandler.test.js
```diff
@@ -99,9 +102,56 @@ describe('OlMeasurementHandler', () => {
...
+			const environmentSpy = spyOn(environmentServiceMock, 'isTouch');
+			const map =  setupMap();
+
+			classUnderTest.activate(map);
+			expect(environmentSpy).toHaveBeenCalled();			
+		});
 
+		it('isTouch() resolves in higher snapTolerance', () => {
+			const classUnderTest = new OlMeasurementHandler();
+			environmentServiceMock.isTouch = () => true;
```
taulinger, line 142: It's often better to use a spy  to define returned values:
`spyOn(environmentServiceMock, 'isTouch').and.returnValue(true);`

Then we can also test for call parameters and call counts if needed...

taulinger (approved):

PR #102, opened 2021-02-23
Refactor consolidate i18n per component

thiloSchlemmer:


taulinger:
There's another convention we didn't mention yesterday (sorry 😥)
`//the first part of the snake_case key should be the name of the related module`
which is important, cause the TranslationService acts as a global registry and we have to avoid key collisions.
The second snake_case_value should be the name of the particular component package.
This means in our case:
- `olMapProvider -> map_olMap_...`
- `layerManagerProvider ->  map_layerManager_..`

We also need an extra provider for `infoButton`

The keys in `contextMenue.provider.js`  have also to be adjusted...

thiloSchlemmer:
a provider for zoomButtons follows..

thiloSchlemmer:
some minor fixes in naming the keys and introcing zoomButtons.provider.  

extend the tests with a check on the translation-map-size

taulinger (approved):

PR #108, opened 2021-02-24
Add possibiltiy to observe state changes of the store

taulinger:
...independently from using a `BaElement` component.

For an example have a look at `measurement.observer.`.
An OlLayerHandler will register an observer for the store the same way . 
In this case, the store has to be injected from the `StoreService`.


thiloSchlemmer (approved):
Got it

PR #109, opened 2021-02-24
Add possibiltiy to observe discrete state changes

taulinger:
... within a BaElement.

Example:

If the state of an element extracted from die global store looks like: 

`{ foo:bar, some:42 }`

...and we want to be notified about changes individually, we can now register an observer:
```
observe('foo', (value) => { console.log('foo changed: ' value) });
observe('some', (value) => { console.log('some changed: ' value)});
```

### DIFF
--- a/src/modules/map/store/measurement.observer.js
+++ b/src/modules/map/store/measurement.observer.js
@@ -1,0 +1,25 @@
+import { observe } from '../../../utils/storeUtils';
+import { addLayer, removeLayer } from './layers.action';
+
+/**
+ * Id of the layer used for measurement interaction
+ */
+export const MEASUREMENT_LAYER_ID = 'measurement_layer';
+
+export const registerMeasurementObserver = (store) => {
+
+	const extract = (state) => {
+		return state.measurement.active;
+
+	};
+	const onChange = (changedState) => {
+		if (changedState) {
+			addLayer(MEASUREMENT_LAYER_ID, { constraints: { hidden: true, alwaysTop: true } });
+		}
+		else {
+			removeLayer(MEASUREMENT_LAYER_ID);
+		}
+	};
+
+	observe(store, extract, onChange);
+};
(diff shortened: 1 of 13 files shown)


src/modules/map/store/measurement.observer.js
```diff
@@ -0,0 +1,25 @@
+import { observe } from '../../../utils/storeUtils';
+import { addLayer, removeLayer } from './layers.action';
+
+/**
+ * Id of the layer used for measurement interaction
+ */
+export const MEASUREMENT_LAYER_ID = 'measurement_layer';
+
+export const registerMeasurementObserver = (store) => {
```
thiloSchlemmer, line 9: to mark this method as a implementation of a recipe or template for store-observer, i would suggest to rename the method to 'register'. This should help to identify and multiply the necessary steps to create new store-obsservers, similar to our usage of i18n-providers

src/services/StoreService.js
```diff
@@ -8,6 +8,7 @@ import { layersReducer } from '../modules/map/store/layers.reducer';
 import { mapContextMenuReducer } from '../modules/map/store/mapContextMenu.reducer';
 import ReduxQuerySync from 'redux-query-sync';
 import { measurementReducer } from '../modules/map/store/measurement.reducer';
+import { registerMeasurementObserver as registerMeasurementObserver } from '../modules/map/store/measurement.observer';
```
thiloSchlemmer, line 11: with the renaming of the method in the measurement.observer-implementation it would change to 
`import { register as registerMeasurementObserver } from...`

thiloSchlemmer:
so...a possible Non-BaElement simply implements the above mentioned steps to observe a state:
```
const extract = (state) => {...};
const onChange = (changedState) => {...};
observe(store, extract, onChange);
```
...and off you go, right?

taulinger:
Observations for state changes outside BaElements is done in #108. I will implement your suggestions there.

This PDR is forked from 108 and gives the possibility to observe changes of a single portion of the state of a BaElement
from **inside** the element.

taulinger:
... so we don't have to detect the updated/changed field within  `onStateChanged`

EnBakir (commented):
Nice to have a possibility to observe state changes. Well done :+1:

Issue #110, opened 2021-02-25
Measurement-Overlays visible after Measurement-Layer is toggled invisible


thiloSchlemmer:
if the measurement-layer changes the visibility or opacity the measurement-related overlays does not change.

thiloSchlemmer:
Issue is fixed.

PR #117, opened 2021-02-26
add changeListeners for measurement-layer

thiloSchlemmer:
add changeListener for measurement-layer to bind layer-visibility and opacity to overlay-visibility/opacity 

fix for #110

the current fix interpretes the measurement-layer as internal object (OlMeasurementHandler is stateful) and therefore using listeners to react on changes of the layer.

But... it breaks the design (use the StateStore as the single point of truth)...so a implementaion with a observer is more appropriate.

### DIFF
--- a/src/modules/map/components/olMap/handler/measure/OlMeasurementHandler.js
+++ b/src/modules/map/components/olMap/handler/measure/OlMeasurementHandler.js
@@ -31,38 +31,51 @@ export class OlMeasurementHandler extends OlLayerHandler {
 		this._mapService = MapService;
 		this._environmentService = EnvironmentService;
 		this._vectorLayer = null;
-		this._draw = false;			
-		this._activeSketch = null;		
+		this._draw = false;
+		this._activeSketch = null;
 		this._helpTooltip;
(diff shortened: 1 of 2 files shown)


thiloSchlemmer:
To compare the version with changeListeners to a version with a StateStore-Observer, i created the branch [fix_measure-overlay_observe-visibility](https://github.com/ldbv-by/bav4-nomigration/tree/fix_measure-overlay_observe-visibility)

the main differences of the different approaches are:

- version with changeListeners 
   - is more lightweight, the behavior is binded to a internal object (the - on activate - created layer)
   - easier to test
- version with observer
   - needs an injected StoreService to build an observer
   - tests more complex, needs .reducer and .actions to create and record state-changes

taulinger (approved):
Nice comparison of both approaches!!
I don't want to decide which one to prefer in this the case...

Note:
In case of needing to be informed 
**after** the **olLayer** has been removed from the map, 
there could be a difference: The observer does not guarantee 
to be called after the olMap.

thiloSchlemmer:
Thx...so for the record: i stick with the changeListener-approach

PR #121, opened 2021-03-01
Pointer and map reducer

taulinger:
Add 
1. a reducer for pointer based interactions of a user with the map
2. a reducer for interactions with the map on a higher abstraction level, like moving the map

So we are able now to
- switch off geolocation tracking when the user is dragging the map
- refactor the trigger mechanism of the context menu


thiloSchlemmer (approved):
Nice enhancement for our interactions with the map.

PR #123, opened 2021-03-02
Add eslint no-var & prefer-const rule

taulinger:
see:
- https://eslint.org/docs/rules/no-var
- https://eslint.org/docs/rules/prefer-const


thiloSchlemmer (approved):

thiloSchlemmer (approved):
Cool enhancement! Improves our code-quality after refactorings (let-> const)

PR #134, opened 2021-03-08
Add information about readiness of injector

taulinger:
When injector is marked as ready, all dependencies are registered and resolvable.


thiloSchlemmer (approved):
Nice enhancement!

PR #138, opened 2021-03-09
modify measurement-feature after drawing

thiloSchlemmer:


taulinger:
If the line contains only two points, the starting point cannot be moved

taulinger (commented):
If the line contains only two points, the starting point cannot be moved

taulinger (commented):
Sorry for the inconvenience...😏
When I create a new vertex  and subsequently want to delete it by clicking on it,
I have to click twice, the first click is ignored.

thiloSchlemmer:
> Sorry for the inconvenience...😏
> When I create a new vertex and subsequently want to delete it by clicking on it,
> I have to click twice, the first click is ignored.

Solved...but now you have be patient with your click (to delete)  ... moving the mouse quickly causes the singleClick not to be recognized

taulinger (commented):
👍 for the fix

When drawing is finished by a double-click, the interactions switches  automatically  into the 'modify' mode, the modify vertices are always shown.
What do you think, would it be better if we first have to select the geometry before we switch into 'modify' mode?

Touch support is fine now!

thiloSchlemmer:
> When drawing is finished by a double-click, the interactions switches automatically into the 'modify' mode, the modify vertices are always shown.

Yes, that's the adopted behavior from ba3 (the last drawn feature is automatically selected) but adjusted to the focus of this userstory

> What do you think, would it be better if we first have to select the geometry before we switch into 'modify' mode?

This is definitely a field for detailed thoughts. I would investigate the behavior of other apps to choose the best variant for us.
Your suggestion is the default behavior in GIS applications, i guess and would be expected to be so.

taulinger (approved):
OK, let's go for a merge...

Very well done!

PR #147, opened 2021-03-12
Encode/Decode position/layer state to/from URL

taulinger:
Share the current map (position and layers)


src/modules/map/store/GeolocationObserver.js
```diff
@@ -3,29 +3,19 @@ import { $injector } from '../../../injection';
...
-			TranslationService: translationService,
-			CoordinateService: coordinateService,
-			EnvironmentService: environmentService,
-			MapService: mapService
-		}
-			= $injector.inject('TranslationService', 'CoordinateService', 'EnvironmentService', 'MapService');
-		this._translationService = translationService;
-		this._coordinateService = coordinateService;
-		this._environmentService = environmentService;
-		this._mapService = mapService;
```
thiloSchlemmer, line 24: what is the reason to move the service-objects directly to they corresponding usage? Is there a potential performance issue to get every time a service from the injector?

thiloSchlemmer (approved):
Great move forward, to have the url-parameter-handling working

Question of understanding inline

src/modules/map/store/GeolocationObserver.js
taulinger: There aren't  any performance issues to be expected, both approaches use  "call-by-reference".
My goal here was just to separate the unneeded local state from the necessary one in order to minimize side effects.

PR #159, opened 2021-03-17
Dev administration provider

EnBakir:
Provides district and community, loaded from Backend.


src/services/provider/administration.provider.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+export const loadBvvAdministration = async (coordinate3857) => {
+
+	const { HttpService: httpService, ConfigService: configService } = $injector.inject('HttpService', 'ConfigService');
+
+	const url = configService.getValueAsPath('BACKEND_URL') + 'administration/info';
+
+
+	const result = await httpService.fetch(`${url}/${coordinate3857[0]}/${coordinate3857[1]}`, {
+		mode: 'cors'
```
taulinger, line 16: 🤓: Not necessarily needed, "mode:  cors" is the default value

src/services/AdministrationService.js
```diff
@@ -0,0 +1,27 @@
+import { loadBvvAdministration } from './provider/administration.provider';
+import { isCoordinate } from '../utils/checks';
+
```
taulinger, line 3: We need a type definition for `Administration` to document the domain object of this service

src/services/AdministrationService.js
```diff
@@ -0,0 +1,27 @@
...
+export class AdministrationService {
+
+	constructor(administrationProvider = loadBvvAdministration) {
+		this._administrationProvider = administrationProvider;
+	} 
+
+	/**
+     * 
+     * @param {Coordinate} coordinate3857 
+	 * @returns {Object} with community and district as string properties
```
taulinger, line 13: Here we have to reference `Administration`

src/services/AdministrationService.js
```diff
@@ -0,0 +1,36 @@
+import { loadBvvAdministration } from './provider/administration.provider';
+import { isCoordinate } from '../utils/checks';
+
+/**
+ * An async function that provides an object 
+ * with community and district as string properties.
```
taulinger, line 6: No, that's not an async function, it's a plain object with two fields 😉

This description would fit better to line 20.
`@async` can be omitted, it is automatically detected by jsdoc

taulinger:
I took the opportunity and added missing docs on all services and providers

taulinger (approved):

PR #163, opened 2021-03-18
Topics management

taulinger:
- load topics configuration from backend
- initialize app with configured bg layer
- store topics state
- encode/decode topic to/from query parameter


src/services/provider/topics.provider.js
```diff
@@ -1,5 +1,38 @@
...
+			let topic = null;
+			topic = new Topic(
+				definition.id, 
+				definition.label, 
+				definition.description, 
+				definition.defaultBackground, 
+				definition.backgroundLayers, 
+				definition.activatedLayers, 
+				definition.selectedLayers);
+			if (topic.id && topic.label) { //at least the id and the label should be set
```
taulinger: `description` and `backgroundLayers` must also be defined

src/services/provider/topics.provider.js
```diff
@@ -1,5 +1,38 @@
...
+
+
+	const result = await httpService.fetch(`${url}`, { });
+
+	if (result.ok) {
+		const topics = []; 
+		const payload = await result.json();
+		payload.topics.forEach(definition => {
+			let topic = null;
+			topic = new Topic(
```
taulinger, line 20: The order of the constructors argument changed. Now it is:
`id, label, description, backgroundLayers, defaultBackground, activatedLayers, selectedLayers`

src/services/provider/topics.provider.js
EnBakir: Added check of property description with commit https://github.com/ldbv-by/bav4-nomigration/pull/163/commits/94e069a0c21a34660b3c5961ebc5d399e90ef7fb

thiloSchlemmer (approved):

PR #167, opened 2021-03-23
fix #152: recognize layer-properties on load

### LINKED ISSUE #152: Layer visibility toggle in layer manager
Layer manager seems not to take account of a falsy initial layer visibility, the visibility toggle is always on.

http://localhost:8080/?c=691366,5335085&z=12&l=atkis,6f5a389c-4ef3-4b5a-9916-475fd5c5962b&l_v=true,false

Let's wait with a fix until #147  is merged into master.

thiloSchlemmer:
fix issue Layer visibility toggle in layer manager #152 with a recognition
of the visibility-property on load

add test for the issue


taulinger (approved):
Thx

PR #174, opened 2021-03-24
Dev scale line

EnBakir:


taulinger (commented):
Sorry, plans did change...
I think we should render each map implementation its own scale line onto the map,
not within the footer. Currently, only OL can generate a scale line in arbitrary HTML elements.
Therefore we can close #174.

EnBakir:
> Sorry, plans did change...
> I think we should let render each map implementation its own scale line onto the map,
> but not within the footer. Currently, only OL can generate a scale line in arbitrary HTML elements.
> Therefore we can close #174.



> Sorry, plans did change...
> I think we should let render each map implementation its own scale line onto the map,
> but not within the footer. Currently, only OL can generate a scale line in arbitrary HTML elements.
> Therefore we can close #174.

it's ok. at least i could use parts of the css for https://github.com/ldbv-by/bav4-nomigration/pull/178

PR #175, opened 2021-03-24
Rename topics fields again

taulinger:
I think I've got it now 🤪


EnBakir (approved):
Although i have to say that 'GeoR' isn't really intuitive. Maybe sth like 'GeoRsrc' would be a bit better. But approved :+1:

taulinger:
> Although i have to say that 'GeoR' isn't really intuitive. Maybe sth like 'GeoRsrc' would be a bit better. But approved +1

You're right, i will change it to your suggestion

PR #179, opened 2021-03-25
Change context menu content order

EnBakir:
The Product Owner wants another context menu content order, therefore the order of the context menu was reordered.


thiloSchlemmer (approved):

taulinger (approved):
A short description about reason / goal of the refactoring would be helpful 😉

PR #192, opened 2021-04-08
fix calculation and usage of partitionDeltas

thiloSchlemmer:
The calculation of the partitionDeltas for a geometry is now based
on the given resolution ot the mapview, to prevent an overload of
partitionpoints in edge cases.

Add listener to the event 'change:resolution' of map.getView() to
refresh the overlays on zoom interactions while drawing

### DIFF
--- a/src/modules/map/components/olMap/handler/measure/GeometryUtils.js
+++ b/src/modules/map/components/olMap/handler/measure/GeometryUtils.js
@@ -166,7 +166,6 @@ export const getPartitionDelta = (geometry, resolution = 1, calculationHints = {
 		}	
 		partitionLength = partitionLength * stepFactor;
 	}
-	console.log(delta);
 	return delta;
 };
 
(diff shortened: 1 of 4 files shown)


thiloSchlemmer:
> Nothing to adapt in `GeometryUtils.test.js`?

Should be :)  The fixed/updated method was only tested within `OlMeasurementHandler.test.js`. This tests are updated. But moving the most part of the tests to `GeometryUtils.test.js`, would be better.

taulinger:
The slightly decreased test coverage made me think of this...

thiloSchlemmer:
ok, i got it....the decreased test coverage is made by untested branch, resulted from a changed behavior ('update tooltips/overlays also on zooming')  

...and fixed a newly introduced bug ('spamming overlays')

taulinger (approved):
Fine!

PR #200, opened 2021-04-09
Handle internally created vector georesources

taulinger:
- Save vector data generated during runtime e.g. by the user to backend. (see example in `ShowCase.js`)
- Load vector data from backend (example: HTTP://localhost:8080/?l=f_099e1d20-9c3c-11eb-89dd-677f9628b0a0_507d8245-df24-404d-a5c3-bc756604b798)


src/services/FileStorageService.js
```diff
@@ -0,0 +1,117 @@
...
+
+
+/**
+ * Service for persisting and loading ASCII based geodata.
+ * @author taulinger
+ * @interface FileStorageService
+ */
+
+/**
+ * Loads the the content of a file.
```
thiloSchlemmer, line 11: Typo

src/services/GeoResourceService.js
```diff
@@ -82,15 +82,20 @@ export class GeoResourceService {
 
 	/**
 	 * Adds a GeoResoure to the internal cache.
+	 * An existing GeoResource will be replaced by the new one.
+	 * The replacement is done based on the id of the geoRersource.
```
thiloSchlemmer, line 86: Typo 'geoRe(r)source'

thiloSchlemmer (approved):
Great Job! The next big thing...due to the fact that this PR is more the nuclear power-plant then the bicycle-shed, i found only bicycle-related issues :)  -> see inline comments

Issue #215, opened 2021-04-14
Layer Manager...


taulinger:
should update the name of a layer when label property changed ( observes changes on `layer.label`)

PR #217, opened 2021-04-15
fix behavior of LayerItem on store-changes

thiloSchlemmer:
fix behavior of LayerItem and LayerManager, if ModifyableLayerProperties
are changed via modifyLayer

add tests for modifyLayer with valid ModifyableLayerProperties 

#215


taulinger (approved):
Thx 👌

PR #218, opened 2021-04-15
Rename i18n

EnBakir:
Changed base layer info label and added fallback content


taulinger (approved):
Good idea to provide a fallback.

PR #219, opened 2021-04-17
Add network related state

taulinger:
- adds a network-related slice of state
- intercepts fetch calls to update networks fetching property
- animates BA logo while fetching
- remove double 'it' expressions in tests

### DIFF
--- a/src/services/HttpService.js
+++ b/src/services/HttpService.js
@@ -1,3 +1,5 @@
+import { setFetching } from '../store/network.action';
+
 /**
  * @class
  * @author aul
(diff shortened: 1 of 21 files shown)


taulinger:
Build shows overall decreased coverage due to poor coverage score of `Autocomplete.js` (https://coveralls.io/jobs/78819405)
Must be an error, did not change anything there. Normally `Autocomplete.js`  has around 97%.

src/services/HttpService.js
```diff
@@ -57,7 +64,7 @@ export class HttpService {
 	 */
 	async post(resource, data, contentType, options = {}) {
 		const fetchOptions = {
-			mode: 'cors',
+			mode: HttpService.DefaultRequestMode,
 			method: 'POST',
```
thiloSchlemmer, line 68: To be consequent, the request method-types could also be extracted to private CONST

thiloSchlemmer (approved):
Good job.... only tiny remark found

src/services/HttpService.js
taulinger: Need more explanation...

taulinger: Probably I should name the static getter `DEFAULT_REQUEST_MODE` to indicate it has public access?

PR #220, opened 2021-04-19
enable multiple measurements 

thiloSchlemmer:
enable multiple measurements in one session (active -> deactivate)  and display the statistics of currently measured geometries or selected Geometries in Toolbox


taulinger (commented):
Code looks good 💪
but my UX was not the best. Maybe it was too late, but I didn't get the idea when
I have to click the activate or delete button for example...

thiloSchlemmer:
OK...i modified the behavior of activation. Nö measurement Starts after click in toolbar and deactivates, If tool changes.... Try it Out

taulinger (approved):
@thiloSchlemmer @alsturm Very nice!

PR #222, opened 2021-04-20
introduce UnitsService

thiloSchlemmer:
as enhancement and refactoring for two utility-methods, this PR introduce a service to format distance- and area-values with their respective unit.

This refactoring is needed for future implementations, where such values have to displayed by arbitrary components with the same formatting rule.

### DIFF
--- a/src/injection/config.js
+++ b/src/injection/config.js
@@ -6,6 +6,7 @@ import { ProcessEnvConfigService } from '../services/ProcessEnvConfigService';
 import { HttpService } from '../services/HttpService';
 import { TranslationService } from '../services/TranslationService';
 import { ShareService } from '../services/ShareService';
+import { UnitsService } from '../services/UnitsService';
 import { GeoResourceService } from '../services/GeoResourceService';
 import { AltitudeService } from '../services/AltitudeService'; 
 import { UrlService } from '../services/UrlService';
@@ -31,6 +32,7 @@ $injector
 	.registerSingleton('AltitudeService', new AltitudeService())
 	.registerSingleton('SearchResultProviderService', new SearchResultProviderService())
 	.registerSingleton('ShareService', new ShareService())
+	.registerSingleton('UnitsService', new UnitsService())
 	.register('FileStorageService', BvvFileStorageService)
 	.register('UrlService', UrlService)
 	.registerSingleton('AdministrationService', new AdministrationService())
--- a/src/injection/config.js
+++ b/src/injection/config.js
@@ -32,7 +32,7 @@ $injector
 	.registerSingleton('AltitudeService', new AltitudeService())
 	.registerSingleton('SearchResultProviderService', new SearchResultProviderService())
 	.registerSingleton('ShareService', new ShareService())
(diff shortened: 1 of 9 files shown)


test/modules/map/components/olMap/handler/measure/MeasurementOverlay.test.js
```diff
@@ -10,7 +12,10 @@ describe('MeasurementOverlay', () => {
 
 	beforeEach(async () => {
 		TestUtils.setupStoreAndDi({});			
-
+		$injector.registerSingleton('ConfigService', {
+			getValue: () => { }
+		});
+		$injector.registerSingleton('UnitsService', new UnitsService());
```
taulinger, line 18: We only  use mocked services  in tests, because we want to focus just on the logic of the code under test and exclude foreign classes and their complexity

src/services/UnitsService.js
```diff
@@ -0,0 +1,81 @@
...
+const KILOMETER_IN_METERS = 1000;
+const SQUAREDKILOMETER_IN_SQUAREDMETERS = 1000000;
+const HEKTAR_IN_SQUAREDMETERS = 10000;
+const PROVIDER_METRIC = {
+	/**
+    * Appends the metric unit of distance to the specified number
+    * @param {number} distance 
+    * @returns {String} the formatted value 
+    */
+	distance:(distance) => {
```
taulinger, line 12: I suggest using `distance(distance) {` as method definition (in this case)

src/services/UnitsService.js
```diff
@@ -0,0 +1,81 @@
+import { $injector } from '../injection';
+
+const KILOMETER_IN_METERS = 1000;
+const SQUAREDKILOMETER_IN_SQUAREDMETERS = 1000000;
+const HEKTAR_IN_SQUAREDMETERS = 10000;
+const PROVIDER_METRIC = {
```
taulinger, line 6: Let's name "private"  static properties in a Pascal_Snake_Case  manner: `Kilometers_In_Meters`.
UPPER_SNAKE_CASE should be reserved for "public" static fields

thiloSchlemmer:
updated as suggested

src/injection/config.js
```diff
@@ -31,6 +32,7 @@ $injector
 	.registerSingleton('AltitudeService', new AltitudeService())
 	.registerSingleton('SearchResultProviderService', new SearchResultProviderService())
 	.registerSingleton('ShareService', new ShareService())
+	.registerSingleton('UnitsService', new UnitsService())
```
taulinger, line 35: Sorry, did not see this the last time... 
Effective stateless services should be registered in "Per Request" scope

thiloSchlemmer: Ah...got it!

taulinger: True, therefore I meant _**Effective** stateless services ..._
Technically it has a state, but it's just a value derived from the Config Service, there no real state needed to be kept up 😌

taulinger (approved):

PR #231, opened 2021-04-27
Search results menu

taulinger:
Implementation of a menu panel that displays search results.


taulinger (commented):
ok, let's implement the debounced calls of #save() first

taulinger:
That's it 😎

thiloSchlemmer (approved):

src/modules/search/components/menu/types/geoResource/GeoResourceResultItem.js
```diff
@@ -0,0 +1,112 @@
...
+		/**
+		 * Uses mouseenter and mouseleave events for adding/removing a preview layer.
+		 * These events are not fired on touch devices, so there's no extra handling needed.
+		 */
+		const onMouseEnter = (result) => {
+			//add a preview layer
+			addLayer(GeoResourceResultItem._tmpLayerId(result.id), { label: result.label, geoResourceId: result.id, constraints: { hidden: true, alwaysTop: true } });
+		};
+		const onMouseLeave = (result) => {
+			//remove the preview layer
```
thiloSchlemmer, line 68: a more philosophical question of our coding style, then a real proposal for change, which @alsturm and i stumbled over in the last changes of the layermanager...

Do we prefer smallest helper methods and expressions with a meaning name, instead of the pure implementation with a meaningful comment.

In this case:
`const removePreviewLayer = () => removeLayer(GeoResourceResultItem._tmpLayerId(result.id));`
instead of:
`//remove the preview layer` 
`removeLayer(GeoResourceResultItem._tmpLayerId(result.id));`

src/modules/search/components/menu/types/geoResource/GeoResourceResultsPanel.js
```diff
@@ -0,0 +1,109 @@
...
+import { requestData } from '../resultPanelUtils';
+import css from './geoResourceResultsPanel.css';
+
+
+/**
+ * Displays geoResource search results.
+ * @class
+ * @author taulinger
+ */
+export class GeoResouceResultsPanel extends BaElement {
```
thiloSchlemmer, line 15: can GeoResourceResultsPanel benefit from the css-definitions of AbstractContentPanel?

src/modules/search/components/menu/types/location/LocationResultsPanel.js
```diff
@@ -0,0 +1,108 @@
...
+			async (term) => {
+				this._locationSearchResults = await requestData(term, searchResultProvider, LocationResultsPanel.Min_Query_Length);
+				this.render();
+			});
+
+		this.observe('term', term => requestLocationDataAndUpdateViewHandler(term), true);
+	}
+
+	onStateChanged() {
+		//we we do nothing here, because we will call #render() manually after search results are available
```
thiloSchlemmer, line 44: typo

src/services/GeoResourceService.js
```diff
@@ -7,6 +7,7 @@
  * @typedef {function():(Array<geoResource>)} georesourceProvider
```
thiloSchlemmer, line 7: Typo: 'that that'

thiloSchlemmer:
see inline comments

src/modules/search/components/menu/types/geoResource/GeoResourceResultItem.js
taulinger: I don't think we should use  a helper method just to make clear what's going on, instead we  can use a meaningful name when we import the action 
`import { foo as bar } ... `

thiloSchlemmer (approved):

src/modules/search/components/menu/types/geoResource/GeoResourceResultItem.js
taulinger: Hm, we should 'allow' both approaches, the intention should be clear...

PR #241, opened 2021-04-30
measurement -  persist data

thiloSchlemmer:
after successful creating measurement-data, the data should be converted/promoted from session-data to full-qualified georesource-data


taulinger (commented):
ok, let's implement the debounced calls of #save() first

taulinger (commented):
Almost done!
- we need a test on the amount of the debounce delay
- I think we can risk it to decrease the debounce delay from 2000 to 1000ms

taulinger (commented):
Sorry, one last thing...

Until now, we do not log stacktraces on the console, 
so in our case we have to change the warn statements to something like this:
`console.warn('Could not store content initially:', error.message);		`

taulinger (approved):
Let's go!

PR #245, opened 2021-05-05
update LayerManager for hidden layer

thiloSchlemmer:
update LayerManager to filter layers with 'hidden'-constraint evaluated
to `true`, to prevent rendering of related information. This behavior is needed
for instance for session-layers of tools.

### DIFF
--- a/test/modules/map/components/layerManager/LayerManager.test.js
+++ b/test/modules/map/components/layerManager/LayerManager.test.js
@@ -67,7 +67,7 @@ describe('LayerManager', () => {
 			expect(toggleElement.checked).toBeFalse();
 		});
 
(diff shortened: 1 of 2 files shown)


test/modules/map/components/layerManager/LayerManager.test.js
```diff
@@ -67,6 +67,27 @@ describe('LayerManager', () => {
 			expect(toggleElement.checked).toBeFalse();
 		});
 
+		it('with one hidden of two layer displays one layer item', async () => {
```
taulinger, line 70: Just as suggestions for a better description:
`it('displays one out of two layers - one is hidden'`
or
`it('takes hidden layers into account'`

taulinger (approved):

PR #250, opened 2021-05-07
enhance styling of measurement values and units

thiloSchlemmer:
splitting measurement-values and -units, to create a better access for
stylings with text-alignments seperated for measurement-values
and -units

update styles by seperating them for measurement-values and -units

add test for changed layout-behavior

### DIFF
--- a/src/modules/toolbox/components/measureToolContent/MeasureToolContent.js
+++ b/src/modules/toolbox/components/measureToolContent/MeasureToolContent.js
@@ -49,8 +49,17 @@ export class MeasureToolContent extends BaElement {
 			remove();
 		};
 
-		const formattedDistance = this._unitsService.formatDistance(statistic.length, 2);
+		const buildPackage = (measurement) => {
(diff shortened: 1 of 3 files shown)


src/modules/toolbox/components/measureToolContent/MeasureToolContent.js
```diff
@@ -76,9 +84,8 @@ export class MeasureToolContent extends BaElement {
 						<span>
 							${translate('toolbox_measureTool_stats_area')}:		
 						</span>						
-						<span class='prime-text'>
-								${unsafeHTML(formattedArea)}
-						</span>
+						<span class='prime-text-value'>${formattedAreaPackage.value}</span>
+						<span class='prime-text-unit'>${unsafeHTML(formattedAreaPackage.unit)}</span>
```
taulinger, line 88: Do we really need to use `unsafeHTML` here?

taulinger (approved):

PR #257, opened 2021-05-10
update add/remove-behavior of measurement-layer

thiloSchlemmer:
update add/remove-behavior of measurement-layer to get a better integration with the layermanager:

- converts a (temporary) session-layer to a full-qualified (with georesource) layer on deactivate
- imports features of a eventually existing full-qualified layer into a (temporary) session-layer


src/modules/map/components/olMap/handler/measure/OlMeasurementHandler.js
```diff
@@ -380,8 +406,9 @@ export class OlMeasurementHandler extends OlLayerHandler {
 	async _save() {		
 		const options = { featureProjection: 'EPSG:3857', rightHanded: true, decimals: 8 };
 		const format = new KML({ writeStyles: true });
+		const features = this._vectorLayer.getSource().getFeatures();		
+		features.forEach(f => this._overlayManager.saveManualOverlayPosition(f));
 		this._storedContent = format.writeFeatures(this._vectorLayer.getSource().getFeatures(), options);
```
taulinger, line 411: I think it would be useful if we refactor all import/export KML logic into a separate file e.g.`/formats/kml.js`.
In the future, we can add here additional logic, like setting a documents name when exporting
```
 kmlString = kmlString.replace(/<Document>/, `<Document><name>${name}</name>`);
```
Later also possible format files are gpx and geoJson...

taulinger:
Nice proceedings!

We should let start the name of the util files lower case to meet our conventions...

taulinger (approved):
Nice!
Would you just add a short description for this PR?

PR #270, opened 2021-05-18
Dev font

alsturm:
add font Open Sans in normal and bold


src/main.css
```diff
@@ -13,7 +13,7 @@
     --border-radius-button: 2.5em;
     --border-radius-button-b: .5em;
     
-    --font-family: "Helvetica Neue",Helvetica,Arial,sans-serif;
+    --font-family: 'Open Sans', sans-seri;
```
taulinger, line 16: Shouldn't it be "sans-serif"?

taulinger (changes requested):

src/main.css
alsturm: copy paste fail fixed...

taulinger (approved):

PR #285, opened 2021-05-25
restore style for layers

thiloSchlemmer:
while converting a session-layer to a full-qualified (with corresponding georesource) layer, a style must be applied

the style could be determined by the relation of feature-Id and tool-Id.


taulinger:
Almost done!

The StyleService needs another method, something like `isStyleRequired()`  (returns a Boolean). Currently I  misuse `_detectStyleType()` for that purpose (see VectorImportService L.103)

taulinger:
Btw: `OverlayStyle.js` seems to have a poor branch coverage (https://coveralls.io/builds/40453440/source?filename=src%2Fmodules%2Fmap%2Fcomponents%2FolMap%2FOverlayStyle.js)

taulinger:
Found a strange behavior: When resizing the browser window the layer manager visually  "loses" the measurement layer  🤔

thiloSchlemmer:
> isStyleRequired

for the current implementation and our current behavior of applying styles, it's sufficient to create a wrapper-method for `_detectStyleType()`, is'nt it? For the moment there are no further elaborate styling requirements defined....if so, we can adopt this wrapper or use a other implementation.

PR #300, opened 2021-06-07
add: new checkbox component

alsturm:


taulinger (commented):
Nice component!
After fixing the layout on android, I would suggest adding an overview comment on top of the class mentioning attributes and properties like in `Button.js`

src/modules/utils/components/showCase/ShowCase.js
```diff
@@ -135,6 +135,11 @@ export class ShowCase extends BaElement {
 			</div>
 			<p>Toggle-Button</p>
 			<div class='toggle' style="display: flex;justify-content: flex-start;"><ba-toggle id='toggle' title="Toggle" @toggle=${onToggle}><span>Toggle me!</span></ba-toggle></div>
+			<p>Checkbox</p>
+			<div><ba-checkbox id='checkbox'  title="checkbox tible" @toggle=${onToggle}><span>checkbox</span></ba-checkbox></div>
+			<div><ba-checkbox id='checkbox' checked=true title="checkbox tible" @toggle=${onToggle}><span>checkbox checked</span></ba-checkbox></div>
+			<div><ba-checkbox id='checkbox' disabled=true title="checkbox tible" @toggle=${onToggle}><span>checkbox disabled</span></ba-checkbox></div>
+			<div><ba-checkbox id='checkbox' checked=true disabled=true title="checkbox tible" @toggle=${onToggle}><span>checkbox checked disabled</span></ba-checkbox></div>
```
taulinger, line 142: always the same id is used

alsturm: removed ids

taulinger (approved):
👍

PR #311, opened 2021-06-16
share a measurement from the measurement tool

thiloSchlemmer:
enable the measureToolContent to publish a link with the current state of the measurement as:

- readonly link; the content of the measurement is not editable and is at the same state at any time
- writeable link; the content of the measurement is editable and can be updated by any user which have the shared link


taulinger (approved):
@thiloSchlemmer @alsturm Very nice,
would suggest just some changes to comments...
After that let's go for a merge!

taulinger:
fixed the ba-button problem.

taulinger:
https://github.com/ldbv-by/bav4-nomigration/blob/defd7a3c713f780ad0e440c09e92269397dd8b48/test/modules/toolbox/components/measureToolContent/MeasureToolContent.test.js#L148

Maybe we should also test alls ba-buttons for the expected value of the attribute "label"?

thiloSchlemmer:
Similar to `it('displays the finish-button', async () => {...` but with detailed attribute-expectations?

taulinger:
> Similar to:
> 
> https://github.com/ldbv-by/bav4-nomigration/blob/defd7a3c713f780ad0e440c09e92269397dd8b48/test/modules/toolbox/components/measureToolContent/MeasureToolContent.test.js#L93
> 
> ... but with detailed attribute-expectations?

 Something like
 `expect(element.getAttribute('label').toBe('toolbox_measureTool_delete_measure')`

src/modules/toolbox/components/measureToolContent/ShareMeasureDialog.js
```diff
@@ -0,0 +1,93 @@
...
+		const translate = (key) => this._translationService.translate(key);
+		const onCopyUrlToClipBoard = async () => this._copyValueToClipboard(url);
+
+		const getShareApiContent = (useShareApi) => {
+			if (useShareApi) {
+				const onClickWithApi = () => {
+					this._environmentService.getWindow().navigator.share({
+						title: translate('toolbox_measureTool_share_link_title'),
+						url: url
+					}).catch(console.error);
```
taulinger, line 57: `catch` does take a function as an argument.
 It also seems that the test is missing for that case

taulinger (commented):
Almost 😉 done (see inline comment)

taulinger (approved):
🥳

Issue #321, opened 2021-06-23
Missing i18n in header


taulinger:
Tab buttons need i18n also for title attributes

PR #324, opened 2021-06-24
add i18n for tab-titles in header

thiloSchlemmer:
fix Missing i18n in header #321


taulinger (commented):
Can we rename the keys to something like
- header_tab_title_topics
- header_tab_button_topics
and so on ...?

taulinger (approved):
Thx!

PR #348, opened 2021-07-09
Fix Virtual Keyboard problem

### LINKED ISSUE #345: Virtual keyboard distorts UI
On Android, the viewport changes to landscape layout  when the virtual keyboard is opening

taulinger:
fixes #345

### DIFF
--- a/src/modules/header/components/Header.js
+++ b/src/modules/header/components/Header.js
@@ -45,22 +45,22 @@ export class Header extends BaElement {
 
 	createView(state) {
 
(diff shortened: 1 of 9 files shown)


src/modules/header/components/Header.js
```diff
@@ -168,7 +143,7 @@ export class Header extends BaElement {
 				<mask class="header__background">
 				</mask>
```
taulinger, line 144: @alsturm Tag looks strange to me...

alsturm: change to div works (don't know why this is so)

thiloSchlemmer (approved):

Issue #349, opened 2021-07-09
Flaky test(s) in OlMeasurementHandler.test.js


taulinger:
```
absoluteOlMeasurementHandler.test.1703910062.js OlMeasurementHandler when activated over olMap looks for measurement-layer and adds the feature for copy on save FAILED
[INFO] 	Expected null to equal Object({ adminId: null, fileId: 'f_lastId' }).

```

taulinger:
Still there 😭:

Expected object to have properties
	    adminId: null
	Error: Expected object to have properties
	    adminId: null
	    at <Jasmine>
	    at /tmp/_karma_webpack_391152/commons.js:228930:57

taulinger:
Hopefully solved with #385

PR #385, opened 2021-07-27
Extract measure storage logic

thiloSchlemmer:
extract storage logic and state from `OlMeasurementHandler` to `MeasurementStorageHandler` to encapsulate storage-specific behavior and simplify they corresponding usage in `OlMeasurementHandler`

benefit of this refactoring is:

fixing #349 by seperating debounce-triggered operations and read/write-accesses to the store in the testcases


taulinger (commented):
as already discussed ....

taulinger (approved):
Looks good to me 🙂

PR #407, opened 2021-08-18
use ecmascript-compat plugin to check code compatibility

taulinger:
Check code compatibility against our target browsers:
- FF (ESR and Android)
- Chrome (Desktop and Android)
- Edge

Checks against Safari (Desktop and iOS) currently throw an (false positive?) error and are temporarily deactivated.


taulinger:
Not needed anymore, using ESLint 8 and es-check we can now check es11 code compatibility

PR #408, opened 2021-08-19
fix LayerManager i18n

thiloSchlemmer:
fix missing i18n-support in module layerManager with import-statement,
rename i18n-key to follow naming convention


taulinger (approved):
Thx!

PR #432, opened 2021-09-20
BaElement2

taulinger:
Due to our previous experiences with `BaElement` we want to simplify and harmonize our central base component regarding
our work against its state and how to change it. However, the lifecycle of the component has proved itself and will remain unchanged.

The new `MvuElement` is based on the Model-View-Update pattern and contains a functional approach for mutating its state, which is already known from Redux.

An implementation example can be found in `src/modules/commons/components/examples/MvuCounter`.

Additionally, our **base components** have been reworked.
**All _data-related_ bindings are property-based only now**, which simplifies them tremendously.
Like before, _event callbacks_ can be registered both by property and attribute bindings.

Have a look at Checkboxes and Buttons in the `Showcase` to see the changes in action.


src/modules/commons/components/autocomplete/AutocompleteSearch.js
```diff
@@ -166,4 +168,26 @@ export class AutocompleteSearch extends BaElement {
...
+	set onSelect(callback) {
+		this._onSelect = callback;
+	}
+
+	get onSelect() {
+		return this._onSelect;
+	}
+
+	/**
+	 * @property {function} provider - Function that takes a term an returns an array of SearchResults
```
thiloSchlemmer, line 184: typo i guess? (and instead of an)

thiloSchlemmer (dismissed):
Big refurbish! :D 
Nicely done!

thiloSchlemmer (approved):

PR #474, opened 2021-10-12
fix flaky test for 'createAnimateFunction'

### LINKED ISSUE #473: Fix flaky test

thiloSchlemmer:
Provide a fix for flaky tests for the 'createAnimateFunction'

fixes #473 
a fix for case 1 of known errors (negative radius)


thiloSchlemmer:
2 fixes are made:

**switching from implizit to explicit type-conversion** 
instead of `+new Date()` to get the timestamp as number using `Date.now() ` ([see](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Date/now))

**fix timing of framestate-creation**
the timings for the framestates are made before the call of the 'createAnimateFunction'. This could cause negative values for radius calculations.
the fix in the test-setup should now enshure,that the framestate have always bigger values then the start-value in the function (frameState.time>start).


To discuss @taulinger :
should we include this known assertion (frameState.time>start) into the createAnimateFunction itself?
`const elapsed = frameState.time - start;`
`const elapsedRatio = (elapsed >= 0 ? elapsed : 0) / duration;`

PR #487, opened 2021-10-19
add some more example components

thevoiceofthruth:
first semester in the bayernatlas uni create some components + unit tests + styles


taulinger (commented):
Would suggest moving all examples from `commons` to a separate module called `examples`

PR #493, opened 2021-10-21
draw-tool buttons

thiloSchlemmer:
Adding a more adaptive behavior to the draw-tool buttons, similar to to behavior for the measure-tool.


src/modules/map/components/olMap/handler/measure/OlMeasurementHandler.js
```diff
@@ -26,6 +25,7 @@ import { isEmptyLayer } from '../../olMapUtils';
 import { emitNotification, LevelTypes } from '../../../../../../store/notifications/notifications.action';
 import { OlSketchHandler } from '../OlSketchHandler';
 import { MEASUREMENT_LAYER_ID, MEASUREMENT_TOOL_ID } from '../../../../../../plugins/MeasurementPlugin';
+import { simulateMouseEvent } from '../../../../../../../test/modules/map/components/olMap/mapTestUtils';
```
taulinger, line 28: Oops, this is a test-only util function. If we really need it, we have to think about reimplementing (and testing) it in the `src` directory

src/store/draw/draw.action.js
```diff
@@ -80,6 +80,17 @@ export const setType = (type) => {
 	});
 };
 
+/**
+ * Set whether the geometry of a drawing is valid or not.
+ * @function
+ */
+export const setValidGeometry = (isValid) => {
```
taulinger, line 87: Would suggest something like `setGeometryIsValid`

src/store/draw/draw.reducer.js
```diff
@@ -23,6 +24,10 @@ export const initialState = {
 	 * @type {String}
 	 */
 	type: null,
+	/**
+	* @type {boolean}
+	*/
+	validGeometry: false,
```
taulinger, line 30: see comment above

src/modules/map/components/olMap/olMapUtils.js
```diff
@@ -41,9 +44,30 @@ export const registerLongPressListener = (map, callback, delay = 300) => {
...
+ * Method to set the focus back on the current map without any further pointer
+ * moves or clicks from real devices
+ * @param {Map} map the map to focus on
+ */
+export const requestMapFocus = (map) => {
+	const view = map.getView();
+	if (map && view) {
+		const x = view.getCenter()[0];
+		const y = view.getCenter()[1];
+		simulateMouseEvent(map, MapBrowserEventType.CLICK, x, y);
```
taulinger, line 71: Same here.

taulinger (changes requested):
We need to change  `simulateMouseEvent` usage -> see inline comments

taulinger:
BTW: `modules/map/components/olMap/formats kml.js` has a poor branch coverage

src/modules/map/components/olMap/olMapUtils.js
```diff
@@ -41,9 +41,16 @@ export const registerLongPressListener = (map, callback, delay = 300) => {
 	});
 };
 
+
+/**
+ * Checks whether a layer contains features or not.
+ * @param {Layer} layer
+ * @returns {boolean}
+ */
 export const isEmptyLayer = (layer) => {
```
taulinger, line 50: We should rename this function to a more precise name e.g. `hasFeatures`  (OL knows much more layer types than the Vector Layer. What does _empty_ mean in that context?).

I wonder why it should return `true`  for the case of an undefined layer 🤓

taulinger:
Wow, we superpassed a 99% coverage!

taulinger (commented):
Thx! Added one comment regarding `olMapUtils.js `

taulinger (approved):
thx!

PR #509, opened 2021-10-26
extend feature highlighting

taulinger:
Allow highlighting of multiple features and different geometry types.
Add actions for appending and removing features by id.


thiloSchlemmer (approved):
Nice!

PR #528, opened 2021-11-05
Icons management

thiloSchlemmer:
load icons-url from backend


src/services/IconsService.js
```diff
@@ -0,0 +1,35 @@
...
+		return this._icons;
+	}
+
+	all() {
+		if (!this._icons) {
+			console.warn('IconsService not yet initialized');
+			return [];
+		}
+		return this._icons;
+	}
```
taulinger, line 34: Caching for this service is good, but lazy loading should be sufficient here. We only load mandatory data eagerly (georesources, topics). This reduces the number of backend requests on startup.

taulinger:
After thinking for a while, i would like to suggest creating a new module (iconSelect?) containing both the `ImageSelect` component and the `IconService` with its provider. 

Reason: It is not really a common reusable component (in the sense of the ba-button or other low-level components. How many `IconSelect` component instances will we use?) and we do not need the `IconService` on the global level also.

src/services/IconService.js
```diff
@@ -0,0 +1,176 @@
...
+export class IconService {
+
+	constructor(iconProvider = loadBvvIcons, iconColorProvider = getBvvIconColor) {
+		this._iconProvider = iconProvider;
+		this._iconColorProvider = iconColorProvider;
+		this._default = this.createDefault();
+		this._icons = null;
+	}
+
+	createDefault() {
```
taulinger, line 26: Does this method have to be public?

src/services/IconService.js
```diff
@@ -0,0 +1,176 @@
...
+
+	/**
+ *
+   * @param {string} url
+  */
+	decodeColor(url) {
+		return this._iconColorProvider(url);
+	}
+
+	isLocal(iconCandidate) {
```
taulinger, line 95: Does this method have to be public?

src/services/provider/iconUrl.provider.js
```diff
@@ -0,0 +1,22 @@
...
+  * A function that returns a url-string for the specified icon-id and color.
+  *
+  * @typedef {Function} iconUrlProvider
+  * @param {string} id the id of a specific icon
+  * @param {Array<number>} color the rgb-color
+  * @returns {string} the url to the icon
+  */
+
+/**
+ * creates a URL based on Image-Resources from BVV services
```
taulinger, line 12: [C]reates

test/service/IconService.test.js
```diff
@@ -0,0 +1,231 @@
...
+			const icons = await instanceUnderTest.all();
+			const defaultIcon = instanceUnderTest.getDefault();
+
+			expect(icons.length).toBe(3);
+			expect(icons[0]).toEqual(defaultIcon);
+		});
+
+		it('initializes the service with default provider', async () => {
+			const instanceUnderTest = new IconService();
+			expect(instanceUnderTest._iconProvider).toEqual(loadBvvIcons);
```
taulinger, line 51: no test for default  `_iconColorProvider`

taulinger (commented):
Looks good, great progress!
Let's talk about how to handle the "no-backend-available" case before merging.

src/services/IconService.js
thiloSchlemmer: yes, it is used in olStyleUtils

test/service/IconService.test.js
thiloSchlemmer: Thanks

src/services/IconService.js
taulinger: ok

taulinger (approved):
Well done!

PR #536, opened 2021-11-10
update AbstractToolContent to MvuElement

thiloSchlemmer:
as preparation for future work on local states in `DrawToolContent.js` and `MeasureToolContent.js` we update  to MvuElement


taulinger (approved):
Nice!

PR #543, opened 2021-11-18
Fix css lint

alsturm:


taulinger (approved):
Nicely done, also the cleanup of unused files 👍

PR #552, opened 2021-11-26
cache LayerInfo results

thevoiceofthruth:


taulinger (changes requested):
No, we need to cache one LayerInfoResult per geoResourceId 😟

src/modules/layerInfo/services/LayerInfoService.js
```diff
@@ -6,30 +6,36 @@ export class LayerInfoService {
 
 	constructor(provider = loadBvvLayerInfo) {
 		this._provider = provider;
+		this._layerInforesults = new Map();
```
taulinger, line 9: Let's name the field  `_layerInfo[R]esults`

taulinger (approved):
Looks good now, thx

PR #556, opened 2021-11-26
optimize highlight animation duration

taulinger:
It is now guaranteed that the highlight animation runs at least 300ms.

@thiloSchlemmer Thanks for your helpful feedback


thiloSchlemmer (approved):
I think that is currently the best compromise, thanks a lot!

PR #557, opened 2021-11-29
increase delay time

taulinger:
to prevent flaky tests in `OlFeatureInfoHandler.test.js` (see https://github.com/ldbv-by/bav4-nomigration/runs/4350856533)


thiloSchlemmer (approved):
Thanks

PR #569, opened 2021-12-01
move from es9 to es11

taulinger:
We can safely use es10 and es11 language features from now on
- https://caniuse.com/?search=es10
- https://caniuse.com/?search=es11


thiloSchlemmer (approved):
Thx!

PR #571, opened 2021-12-02
disable Share-Button when backend is not available

thiloSchlemmer:
While Backend is not available, Measure- and Draw-Tool should work in a 'Demo'-Mode: drawings and measurements can be created, but not exported or shared, due to technical reasons.

To reflect this behavior, the share-button in the toolbar should be disabled, when backend-connection is not available.


src/modules/toolbox/components/toolBar/ToolBar.js
```diff
@@ -107,7 +119,7 @@ export class ToolBar extends BaElement {
 							${translate('toolbox_toolbar_draw_button')}
 						</div>  					
 					</button>  				               
-					<button  @click="${toggleShareTool}" class="tool-bar__button">
+					<button  @click="${toggleShareTool}" ?disabled=${!backendIsAvailable()} class="tool-bar__button">
```
taulinger, line 122: Let's better use `this._environmentService.isStandalone()`

thiloSchlemmer:
As discussed...not needed. Moreover this behavior is /will be implemented in the Measure-/DrawToolContent

PR #578, opened 2021-12-03
update styling-behavior by using Input-Event

thiloSchlemmer:
In the DrawToolContent the user can change the style of the drawing.
Changes on Color- and Text-Style are made with Input-elements.
Switching to listen to input-events  leads to a more
responsive behavior of the DrawToolContent


taulinger (approved):
Very nice!

PR #582, opened 2021-12-06
prefer async/await

taulinger:
and add corresponding eslint rule


thiloSchlemmer (approved):
Big thanks for this big clean-up :)

PR #584, opened 2021-12-07
fix for drawing/measurement-store

thiloSchlemmer:
Fixing the behavior of MeasurementStorageService, to write the received FileSaveResult to both, draw- and measurement-store.

@taulinger : the fix writes currently in both slices (draw and measure) the same fileSaveResult, should we cut the store at this point into a extra data-slice (i could'nt find a good name ...storage?data? whatsoever...) in the store?

Instead of:  

- store/measurement/fileSaveResult
- store/draw/fileSaveResult

seperated in only:

- store/[data-slice]

### DIFF
--- a/src/modules/map/components/olMap/services/MeasurementStorageService.js
+++ b/src/modules/map/components/olMap/services/MeasurementStorageService.js
@@ -99,7 +99,7 @@ export class MeasurementStorageService {
 			else {
 				try {
 					const fileSaveResult = await fileStorageService.save(null, content, type);
-					setMeasurementFileSaveResult(fileSaveResult);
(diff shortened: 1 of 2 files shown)


taulinger:
> @taulinger : the fix writes currently in both slices (draw and measure) the same fileSaveResult, should we cut the store at this point into an extra data-slice (i could'nt find a good name ...storage?data? whatsoever...) in the store?
> 
> Instead of:
> 
>     * store/measurement/fileSaveResult
> 
>     * store/draw/fileSaveResult
> 
> 
> seperated in only:
> 
>     * store/[data-slice]

Good point, but let's do this in an extra PR

taulinger (approved):

PR #598, opened 2021-12-09
rework event handling for touch devices

taulinger:
regarding touch ("short-press") and long-press events


thiloSchlemmer (approved):
Thx!

PR #599, opened 2021-12-09
add: portrait css, links word break

alsturm:
- in portrait mode the table is desplayed as a colum.

### DIFF
--- a/src/modules/featureInfo/components/FeatureInfoPanel.js
+++ b/src/modules/featureInfo/components/FeatureInfoPanel.js
@@ -13,6 +13,7 @@ import { isTemplateResult } from '../../../utils/checks';
 
 const Update_FeatureInfo_Data = 'update_featureInfo_data';
 export const TEMPORARY_FEATURE_HIGHLIGHT_ID = `highlightedFeatureInfoGeometry_${createUniqueId()}`;
+const Update_IsPortrait = 'update_isPortrait_hasMinWidth';
 
 
 /**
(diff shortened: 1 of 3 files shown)


src/modules/featureInfo/components/FeatureInfoPanel.js
```diff
@@ -13,6 +13,7 @@ import { isTemplateResult } from '../../../utils/checks';
 
 const Update_FeatureInfo_Data = 'update_featureInfo_data';
 export const TEMPORARY_FEATURE_HIGHLIGHT_ID = `highlightedFeatureInfoGeometry_${createUniqueId()}`;
+const Update_IsPortrait = 'update_isPortrait_hasMinWidth';
```
taulinger, line 16: Can we swap line16 with line15?

src/modules/featureInfo/components/FeatureInfoPanel.js
```diff
@@ -31,6 +32,7 @@ export class FeatureInfoPanel extends AbstractMvuContentPanel {
 		this._translationService = TranslationService;
 
 		this.observe(store => store.featureInfo.current, current => this.signal(Update_FeatureInfo_Data, [...current]));
+		this.observe(state => state.media, media => this.signal(Update_IsPortrait, { isPortrait: media.portrait }));
```
taulinger, line 35: Would suggest the same approach as in line34, we broadcast just the value, not an object:
```
this.observe(state => state.media, media => this.signal(Update_IsPortrait, media.portrait ));
```

src/modules/featureInfo/components/FeatureInfoPanel.js
```diff
@@ -40,6 +42,8 @@ export class FeatureInfoPanel extends AbstractMvuContentPanel {
 		switch (type) {
 			case Update_FeatureInfo_Data:
 				return { ...model, featureInfoData: [...data] };
+			case Update_IsPortrait:
+				return { ...model, ...data };
```
taulinger, line 46: Then we would have to adjust this line as follows:
```
return { ...model, isPortrait: data };
```

taulinger (commented):
yeah, nicely done!
(added two suggestions)

PR #605, opened 2021-12-16
prevent unnecessarily triggered save  onDeactivate

thiloSchlemmer:
Fixing unnecessarily triggered calls of _save(), caused by changes of unfinished drawn features.
This happens after drawing a feature (without deselect) and closing the drawTool immediately


taulinger (approved):
Thx!

PR #606, opened 2021-12-16
Drawing: enable Symbol with Text

thiloSchlemmer:
parallel to simple Text-Placements, the draw of symbol should also contain a possible text (Label)


taulinger (approved):
Looks good!

PR #609, opened 2021-12-19
Ruler-style for measurement

thiloSchlemmer:
Adding a better fitting style for measurements, including enhancements on the partition-overlays, to guarranty to have always a clear look on the measurement geometry.


alsturm (approved):

src/modules/map/components/olMap/olGeometryUtils.js
```diff
@@ -203,6 +203,38 @@ export const isVertexOfGeometry = (geometry, vertexCandidate) => {
 	return result ? true : false;
 };
 
+export const moveParallel = (fromPoint, toPoint, distance) => {
```
taulinger, line 206: Would be nice if we have a short description as comment here also

src/modules/map/components/olMap/olGeometryUtils.js
```diff
@@ -203,6 +203,38 @@ export const isVertexOfGeometry = (geometry, vertexCandidate) => {
...
+	];
+	const movedTo = [
+		Math.sin(angle) * distance + toPoint[0],
+		-Math.cos(angle) * distance + toPoint[1]
+	];
+	return new LineString([movedFrom, movedTo]);
+
+};
+
+export const calculatePartitionResidualOfSegments = (geometry, partition) => {
```
taulinger, line 221: Would be nice if we have a short description as comment here also

taulinger (commented):
Very good optimization!

PR #618, opened 2021-12-23
refactor tool handling

taulinger:
and make it possible to change between tools without explicitly closing them.


thiloSchlemmer (approved):
Thanks for the refactoring. Nice simplifying of the toolbox-module.

PR #636, opened 2022-01-11
fix update measure stats in store on multiselect

thiloSchlemmer:
fixes behavior in OlMeasurementHandler, where multiselection of arbitary features (draw- and measure-feature) does not resolve in a update of the statistic-values in the store


taulinger (approved):
Thx!

PR #644, opened 2022-01-17
create olFeatures with attributes

thiloSchlemmer:
To get valid response for featureInfo-requests on local vector-layers, it is necessary to provide name and description-attributes at newly created drawing-features.


taulinger (approved):
Nice!

PR #662, opened 2022-01-26
load GeoResources on-demand

taulinger:
Load non-known GeoResources on-demand. 
For loading a GeoResource in an asynchronous manner the new type ` GeoResourceFuture` is introduced.


thiloSchlemmer (approved):
Wow...big bunch of changes....Thx!

PR #673, opened 2022-02-01
fix draw-interactions

thiloSchlemmer:
Fixing issues with user-interactions while drawing->selecting->drawing...

OlDrawHandler: 

- use _setSelection, setSelection and this._select.getFeatures().clear in correct manner
- introduce _setSelected-Method for calls, where feature is not in vectorsource per ID available

DrawToolContent: 

- add clearDescription to tool.activate-function to reset Description for each new Drawing
- symbolSrc is now created on each drawing-type and call of changeColor

IconService:

- start loading of icons in constructor-call to enable access in getIconResult-method

draw.action/draw.reducer: 

- add clearDescription()

add/update tests


alsturm (commented):
very nice. still one Problem..  

Wehen click at tow overlapping Geometrys both of them are going to be selected...

taulinger (commented):
without checking functionality, I can say that code and test coverage looks good
(geometryInfo.css needs to be linted)

thiloSchlemmer:
> very nice. still one Problem..
> 
> Wehen click at tow overlapping Geometrys both of them are going to be selected...

ahh...thanks a lot, i missed that and will change it ASAP!

alsturm (approved):

PR #679, opened 2022-02-02
add Provider for Metric Units in Bvv-Style

thiloSchlemmer:


taulinger (changes requested):
Let's make the UnitsServie configurable in the same way we have it in other Services, e.g. `MapService`
This means in our case we have to define two provider functions (area, distance) that are set in the constructor of a service. Don't use/extend the ConfigService for that cases 🤓

src/services/provider/units.provider.js
```diff
@@ -0,0 +1,89 @@
...
+* Appends the metric unit of area to the specified number with bvv defined decimals
+* @param {number} area the area value
+* @returns {String} the formatted value
+*/
+export const bvvAreaUnitsProvider = (area) => {
+	const asSquaredKilometer = (areaValue) => {
+		return (Math.round((areaValue / Squaredkilometer_In_Squaredmeters) * 100) / 100).toFixed(3) + ' ' + 'km&sup2;';
+	};
+	const asSquaredMeter = (areaValue) => {
+		return areaValue > 1 ? (Math.round(areaValue * 100) / 100).toFixed(0) + ' ' + 'm&sup2;' : areaValue > 0.5 ? '1 m&sup2;' : '0.5 m&sup2;';
```
taulinger, line 30: line is not fully covered

src/services/provider/units.provider.js
```diff
@@ -0,0 +1,89 @@
...
+		if (decimals) {
+			formatted = (Math.round((distance / Kilometer_In_Meters) * 100) / 100).toFixed(decimals) + ' ' + 'km';
+		}
+		else {
+			formatted = (Math.round((distance / Kilometer_In_Meters) * 100) / 100) + ' ' + 'km';
+		}
+	}
+	else {
+		if (decimals) {
+			formatted = distance !== 0 ? (Math.round(distance * 100) / 100).toFixed(decimals) + ' ' + 'm' : '0 m';
```
taulinger, line 54: line is not fully covered

taulinger (commented):
Looks good now 👍

PR #682, opened 2022-02-07
 add a service for importing vector data

taulinger:
Adds a service for importing vector data both from local sources and URLs.
The service wraps all the necessary stuff which has to be done to create a `VectorGeoResource` both for sync and async
loading.
Based on this, it is a breeze to bring vector data to the map:
```
// create a GeoResource which will be automatically added to the store
const { id, label }  = importService.importVectorDataFromUrl('https://www.geodaten.bayern.de/ba-data/Themen/kml/huetten.kml');
//call the layer action
addLayer(id, { label: label });
```


thiloSchlemmer (dismissed):
Nice!

PR #687, opened 2022-02-09
drag&drop import panel

thiloSchlemmer:
Add  drag&drop-behavior to import Text, URL or Files as VectorResources

### DIFF
--- a/src/modules/dndImport/components/DndImportPanel.js
+++ b/src/modules/dndImport/components/DndImportPanel.js
@@ -11,33 +11,33 @@ const Update_Drag_Active = 'update_drag_active';
  */
 export class DndImportPanel extends MvuElement {
 
-    constructor() {
-        super({
-            isDragActive: false
-        });
-        const { TranslationService } = $injector.inject('TranslationService');
-        this._translationService = TranslationService;
-    }
+	constructor() {
+		super({
+			isDragActive: false
+		});
+		const { TranslationService } = $injector.inject('TranslationService');
+		this._translationService = TranslationService;
+	}
 
-    /**
+	/**
  * @override
  */
-    update(type, data, model) {
-        switch (type) {
-            case Update_Drag_Active:
-                return { ...model, isDragActive: data };
-        }
-    }
+	update(type, data, model) {
+		switch (type) {
+			case Update_Drag_Active:
+				return { ...model, isDragActive: data };
+		}
+	}
 
-    /**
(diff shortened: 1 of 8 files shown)


thiloSchlemmer:
as discussed reaching first steps of implementation (mimick further data-handling with emitting notifications) and is ready for a intermediate review

src/modules/dndImport/components/DndImportPanel.js
```diff
@@ -0,0 +1,155 @@
...
+			stopRedirectAndDefaultHandler(e);
+		};
+		const onDragLeave = (e) => {
+			stopRedirectAndDefaultHandler(e);
+			this.signal(Update_DropZone_Content, null);
+		};
+
+		const onDrop = (e) => {
+			stopRedirectAndDefaultHandler(e);
+			const types = e.dataTransfer.types || [];
```
taulinger, line 74: branch not fully covered

src/modules/dndImport/i18n/featureInfo.provider.js
```diff
@@ -0,0 +1,27 @@
+export const provide = (lang) => {
```
taulinger, line 1: wrong file name

thiloSchlemmer: CopyPasta-Desasta! 😁

taulinger:
> suggestion for the next step (store):
> ### Slice: import
> #### import.action:
> 
>     * function: `importUrl(url)` -> `new EventLike({cdata:null,file:null, url:url})`
> 
>     * function: `importFile(data)` -> `new EventLike({cdata:null,file:fileData, url:null})`
> 
>     * function: `importText(data)` -> `new EventLike({cdata:data,file:null, url:null})`
> 
> 
> #### import.reducer:
> 
>     * IMPORT_ADDED -> store/import/latest -> (payload)->` {cdata:[string|null],file:[string|null], url:[string|null]}`

Wouldn't be two fields for (c)data and URL enough currently? 
But If we would support binary data later, a mime-type field will be mandatory.

taulinger (commented):
a few small changes are needed, see inline comments

taulinger:
Just no to forget: we have to add an e2e test for `<ba-dnd-import-panel>` in `page.spec.test`

taulinger (approved):
Let's go for a merge 🥳

PR #700, opened 2022-02-15
Enhance notifications

thiloSchlemmer:
Remove LevelTypes.CUSTOM to clearify the distinction between a level-based notification and a fixed notification. 

A fixed notification is always a sort of custom notification (content is string or TemplateResult), permanent (without autoclose-time, triggered by user-interaction) and should stay in a position where the current tasks and actions of the user are not interrupted by the notification, but are recognizable. 
Fixed notification are made for optional tasks and decisions, which are not  essential for the current workflow, so a cancelation with any user-interaction is enough as criteria. But content-creators can trigger the cancelation by self with `clearFixedNotification`.

A level-based notification is made for informations which are needed to presented to the user without any need of a following
user-decision (autoclose-time, no user-interaction).


taulinger (commented):
Good rework!
Would suggest designing the `emitFixedNotification` and `emitNotification` action similar to the modal action. The content parameter must be either a plain `String` or a `TemplateResult`.
`Null` should not be needed here, we still have the `clearFixedNotification` action.

taulinger (approved):
Thx!

PR #704, opened 2022-02-16
improve map and layer setup

taulinger:
Provide a global max zoom level property for the map and allow individual min/max zoom levels fo georesources


thiloSchlemmer (approved):
Thx!

PR #709, opened 2022-02-16
fixes #702

### LINKED ISSUE #702: import.action does not set unused field to null

thiloSchlemmer:
sets unused fields to null

updates doc

### DIFF
--- a/src/store/import/import.action.js
+++ b/src/store/import/import.action.js
@@ -4,12 +4,15 @@ import { IMPORT_ADDED } from './import.reducer';
 
 
 /**
- * Request for fitting a map to a geographic extent
- * @typedef {Object} ImportOption
+ * Properties for a import of vectorDataResources
+ * @typedef {Object} ImportProperties
  * @property {String|null} data the data to import
  * @property {String|null} mimeType the mimeType of the data-property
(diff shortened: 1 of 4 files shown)


src/store/import/import.action.js
```diff
@@ -4,13 +4,16 @@ import { IMPORT_ADDED } from './import.reducer';
 
 
 /**
- * Request for fitting a map to a geographic extent
- * @typedef {Object} ImportOption
+ * Properties for a import of vectorDataResources
```
taulinger, line 7: Not necessarily vector data,  could also be  a WMS URL or something else in the future

src/store/import/import.reducer.js
```diff
@@ -5,7 +5,7 @@ export const IMPORT_ADDED = 'import/added';
 export const initialState = {
 
 	/**
-	 * @property {ImportOption|null}
+	 * @property {ImportProperties|null}
```
taulinger, line 8: payload is an  `EventLike<ImportProperties>`

taulinger (commented):
Looks good!

taulinger (approved):
Thx!

PR #710, opened 2022-02-16
georesources: assign optional properties by method

taulinger:
Improves and unifies handling of optional properties for GeoResources and provides a basis for adding further optional properties for all kinds of GeoResources.


thiloSchlemmer (approved):
Nice Rework!

PR #711, opened 2022-02-16
change behaviour of baselayer switcher

### LINKED ISSUE #705: baselayer switcher removes layer even when it has an index > 0
When selecting a base layer, an existing base layer should be removed only when it has an index of 0

alsturm:
fixes #705


taulinger (commented):
Looks good, would just suggest correcting the test description

taulinger:
Obsolete due to #718, but let's keep this branch for a while

PR #712, opened 2022-02-16
fix unintended activation of Dropzone

thiloSchlemmer:
update of the drag&drop-behavior in DndImportPanel, where a
unintended activation of the dropzone by dragged LayerItems is
prevented by checking for emptyness of the dataTransfer.types-Array.
Any other type than the known importTypes (File, text/plain) will
still result in a message being displayed.


taulinger (commented):
Thx, would just suggest correcting the test description

taulinger (approved):
👍

PR #718, opened 2022-02-17
alter BaseLayerSwitcher behavior

### LINKED ISSUE #705: baselayer switcher removes layer even when it has an index > 0

taulinger:
Allow multiple layers for the same GeoResource.
When a layer referencing a base GeoResource is not on index = 0, a second layer for that GeoResource can be added.

Fixes #705


alsturm (approved):

thiloSchlemmer (approved):
Nice

PR #728, opened 2022-02-22
fix flaky DndImportPanel.test

thiloSchlemmer:


thiloSchlemmer:
fix of flaky DndImportPanel.test will be adressed in [dev_rework-import](https://github.com/ldbv-by/bav4-nomigration/pull/730)

PR #744, opened 2022-03-02
LayerManager: additional layer functions

thiloSchlemmer:
additional functions:

- duplicate Layer
- expand all Layer
- collapse all Layer


src/modules/layerManager/i18n/layerManager.provider.js
```diff
@@ -11,7 +11,12 @@ export const provide = (lang) => {
 				layerManager_expand: 'expand properties',
 				layerManager_move_up: 'move Layer up',
 				layerManager_move_down: 'move Layer down',
-				layerManager_remove: 'remove Layer'
```
taulinger, line 14: Would suggest that we name this process consistently "to clone"

src/modules/layerManager/i18n/layerManager.provider.js
```diff
@@ -11,7 +11,12 @@ export const provide = (lang) => {
...
+				layerManager_remove: 'remove Layer',
+				layerManager_duplicate: 'duplicate Layer',
```
taulinger, line 15: clone layer

src/modules/layerManager/i18n/layerManager.provider.js
```diff
@@ -11,7 +11,12 @@ export const provide = (lang) => {
...
+				layerManager_layer_copy: 'Copy',
```
taulinger, line 16: clone

taulinger (commented):
Very nice addition!

src/modules/layerManager/i18n/layerManager.provider.js
thiloSchlemmer: the key `layerManager_layer_copy` is used as suffix for the label of a cloned layer, i guess '[Layername] Copy' is more familiar to the user than '[Layername] Clone'

PR #748, opened 2022-03-03
adjust header labels and initial behavior

taulinger:


thiloSchlemmer:
I like your suggestion...

'catalog' and 'map' are terms which are closer coupled to our implemented functions.

**'Catalog'** implies that you can select something from a thematically ordered list. Which is better than 'Topic'..what action is related to topic? The user have to figure this out first.

**'Map configuration'** is more specific than 'Layers configuration' and implies, that all actions the user can taken, are resulting in a direct change of the map apearance.

taulinger:
Some parts have been realized in the meantime

PR #757, opened 2022-03-08
defaultStyle for VectorLayers

thiloSchlemmer:


src/modules/map/injection/index.js
```diff
@@ -11,13 +11,13 @@ import { OlFeatureInfoHandler } from '../components/olMap/handler/featureInfo/Ol
 
 export const mapModule = ($injector) => {
 	$injector
+		.registerSingleton('StyleService', StyleService)
```
taulinger, line 14: Singleton must be registered with an instance -> `.registerSingleton('StyleService', new StyleService())`

taulinger (changes requested):
Well done 👍,
but currently broken due to falsely registered StyleService

taulinger (approved):
Thx

PR #758, opened 2022-03-08
import vector data from search input field

taulinger:
Import vector data  by supplying cdata or an URL to the search input field


thiloSchlemmer (approved):
Well done!

PR #775, opened 2022-03-15
Default-Style use layer-based color

thiloSchlemmer:
Features without style-information are provided with a default style. 

The current implementation sets a default style with a different color for each feature, regardless of whether the objects all belong to the same layer or not. 

With this enhancement we try to establish a reliable way to set a default style for unstyled features per layer, as it is a standard behavior in desktop gis.

### DIFF
--- a/src/modules/map/components/olMap/handler/draw/OlDrawHandler.js
+++ b/src/modules/map/components/olMap/handler/draw/OlDrawHandler.js
@@ -147,7 +147,7 @@ export class OlDrawHandler extends OlLayerHandler {
 						f.getGeometry().transform('EPSG:' + vgr.srid, 'EPSG:' + this._mapService.getSrid());
 						f.set('srid', this._mapService.getSrid(), true);
 						this._styleService.removeStyle(f, olMap);
-						this._styleService.addStyle(f, olMap);
+						this._styleService.addStyle(f, olMap, layer);
 						layer.getSource().addFeature(f);
 						f.on('change', onFeatureChange);
 					});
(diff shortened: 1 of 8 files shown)


src/modules/map/components/olMap/handler/draw/OlDrawHandler.js
```diff
@@ -147,7 +147,7 @@ export class OlDrawHandler extends OlLayerHandler {
 						f.getGeometry().transform('EPSG:' + vgr.srid, 'EPSG:' + this._mapService.getSrid());
 						f.set('srid', this._mapService.getSrid(), true);
 						this._styleService.removeStyle(f, olMap);
-						this._styleService.addStyle(f, olMap);
+						this._styleService.addStyle(f, olMap, layer);
```
taulinger, line 150: Do we have no test reflecting the StyleService's call?

src/modules/map/components/olMap/handler/measure/OlMeasurementHandler.js
```diff
@@ -128,7 +128,7 @@ export class OlMeasurementHandler extends OlLayerHandler {
 						f.set('srid', this._mapService.getSrid(), true);
 						layer.getSource().addFeature(f);
 						this._styleService.removeStyle(f, olMap);
-						this._styleService.addStyle(f, olMap);
+						this._styleService.addStyle(f, olMap, layer);
```
taulinger, line 131: Do we have no test reflecting the StyleService's call?

taulinger (commented):
Nice enhancement,
it seems that the OlDrawHandler and the OlMeasurementHandler test do not 
verify the arguments of the StyleService call. Should we add this?

taulinger (approved):
THX

PR #784, opened 2022-03-17
MeasureToolContent: add test-ids for acceptance-tests

thiloSchlemmer:
creating test ids on elements for:

- distance-value and -unit
- area-value and -unit
- action-buttons ('startNew','remove', 'finish')


taulinger (approved):
Nice

PR #786, opened 2022-03-17
optionally load external config

taulinger:
Config parameters are set within the `.env` file and loaded from the `process.env` object by the ProcessEnvConfigService.

Additionally, config parameters can be also set now via a `window.config` object, appointed within a script provided from `/config.js`.


thiloSchlemmer (approved):
Nice!

PR #798, opened 2022-03-21
fix: 'alwaysTop'  does not have an impact when adding layers

### LINKED ISSUE #787: Geolocation-Marker disappears
Precondition: Geolocation is active and the geolocation-marker is displayed on the map
Action:  Adding a new (fully opaque) layer
Cause: Geolocation-marker disappears

additional note: deactivating the currently added layer results in a redisplay of the geolocation, so the layer for the gelocation seems to be there but underneath the added layer.

taulinger:
fixes #787


thiloSchlemmer (approved):
Thx

PR #801, opened 2022-03-22
Fix measurestate while drawing in touch-environments

### LINKED ISSUE #799: Measure- & Draw-State is broken in Touch-Environment
While drawing measurement- or drawing-features with touch-devices the state of drawing is not calculated.

The action-Buttons in the measurementToolContent have no effect and the ui seeems to be frozen.

thiloSchlemmer:
fixes #799


taulinger (approved):
I Don’t Like Mondays 🎵

PR #808, opened 2022-03-23
Enhancement ruler style pixel ratio

thiloSchlemmer:
on displays with higher resolution the line-segments for the subticks are not properly moved parallel.

This branch takes the pixelRatio into account and creates a correct line-segment.


taulinger (approved):
👍

PR #819, opened 2022-03-24
fix failing test id generation

### LINKED ISSUE #818: test id generations fails on first call of #render
Reason: when #render() is called the first time, the window object is not yet available

taulinger:
fixes #818 by removing usage of global window property 'ba_enableTestIds''


thiloSchlemmer (approved):
THX

PR #832, opened 2022-03-28
use  loadstart/loadend events from ol.map

taulinger:
use new loadstart/loadend events from ol.map (available since v6.14)


thiloSchlemmer (approved):
Nice!

Issue #840, opened 2022-03-29
fix LayerItem


taulinger:
- explicitly set geoResourcesId for cloning
- use `null` as default layer object

taulinger:
fixed with #842

PR #842, opened 2022-03-29
fix LayerItem

thiloSchlemmer:
correct use of geoResourceId

explicitly set geoResourcesId for cloning

default layer-value in model is null

fixes LayerItem #840

### DIFF
--- a/src/modules/layerManager/components/LayerItem.js
+++ b/src/modules/layerManager/components/LayerItem.js
@@ -1,4 +1,4 @@
-import { html } from 'lit-html';
+import { html, nothing } from 'lit-html';
 import css from './layerItem.css';
 import { $injector } from '../../../injection';
 import { classMap } from 'lit-html/directives/class-map.js';
@@ -35,7 +35,7 @@ export class LayerItem extends AbstractMvuContentPanel {
 
 	constructor() {
 		super({
-			layer: { id: '', label: '', visible: true, collapsed: true, opacity: 1 }
+			layer: null
 		});
 		const { TranslationService } = $injector.inject('TranslationService');
 		this._translationService = TranslationService;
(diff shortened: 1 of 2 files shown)


src/modules/layerManager/components/LayerItem.js
```diff
@@ -126,7 +138,7 @@ export class LayerItem extends AbstractMvuContentPanel {
 
 		const duplicateLayer = () => {
 			//state store change -> implicit call of #render()
-			addLayer(`${layer.geoResourceId}_${createUniqueId()}`, { ...layer, label: `${layer.label} (${translate('layerManager_layer_copy')})`, zIndex: layer.zIndex + 1 });
+			addLayer(`${layer.geoResourceId}_${createUniqueId()}`, { ...layer, geoResourceId: layer.geoResourceId, label: `${layer.label} (${translate('layerManager_layer_copy')})`, zIndex: layer.zIndex + 1 });
```
taulinger, line 141: We should verify having a layer object containing a geoResourceId property in the store

src/modules/layerManager/components/LayerItem.js
```diff
@@ -89,6 +97,10 @@ export class LayerItem extends AbstractMvuContentPanel {
 	createView(model) {
 		const translate = (key) => this._translationService.translate(key);
 		const { layer } = model;
+
+		if (layer == null) {
+			return nothing;
+		}
```
taulinger, line 103: We should add a test for  this case also

taulinger (commented):
Thx, would suggest two more tests...

src/modules/layerManager/components/LayerItem.js
taulinger: `if(!layer)` or `layer === null ` 😉

taulinger (approved):
Nice, thx!

PR #863, opened 2022-04-02
change feature highlight for long-press results

taulinger:
Let's use the same symbol we use for a successful feature info indication
grafik


thiloSchlemmer (approved):
Nice

PR #875, opened 2022-04-04
add query parameter "mid"

taulinger:
defining the initially shown tab of the main menu.
Example: http://localhost:8080/?mid=1 opens the "Map" tab


thiloSchlemmer (approved):
Nice!

PR #882, opened 2022-04-06
fix updatestyle-behavior

### LINKED ISSUE #881: changing color in drawTool broken
Changing color in drawTool does not instantly update the color/style of the selected feature.

Only the `SelectionStyle` of the feature reflects the update instantly.

thiloSchlemmer:
Corrects the updateStyle-behavior by merging correctly the new, changed style, with the current style-array.
This results in a replacement of the first entry in the style-array.

update test; reflecting the changed expectations setStyle should be called with [newStyle, oldStyle2]

fixes #881


taulinger (approved):
Thx

PR #885, opened 2022-04-06
provide also the geoResourceId as a property of the ol layer

taulinger:
... as a starting point for further development e.g. loading existing features within an OlLayerHandler


thiloSchlemmer (approved):
THX!

PR #894, opened 2022-04-07
Fix edit of shared drawings

thiloSchlemmer:
fixes bug where drawings, which are shared as "share to collaborate", were not editable anymore


taulinger (approved):
Yeah!

PR #895, opened 2022-04-07
fix missing slash replacement and query string encoding

taulinger:
affecting all SearchResult providers.

Credits @AlanNaum


thiloSchlemmer (approved):
Nice!

PR #903, opened 2022-04-08
improve deep comparision approach

taulinger:
make `equal` generally valid including  `{ a:0, b:1 } === { b:1, a:0 } `


thiloSchlemmer (approved):
Nice

PR #905, opened 2022-04-08
enhancement; clear input value, before import

### LINKED ISSUE #904: import file; enable import of same file
in Chrome: 
after successfully import a file, selecting the same file with file-chooser does not results in a change-event and therefore prevents a second import.

thiloSchlemmer:
after successfully import a file, selecting the same file with file-chooser does not results in a change-event and therefore prevents a second import.

Clearing the value of the input-element triggers the change-event in the case of the same file also.

fixes #904

### DIFF
--- a/src/modules/toolbox/components/importToolContent/ImportToolContent.js
+++ b/src/modules/toolbox/components/importToolContent/ImportToolContent.js
@@ -27,10 +27,13 @@ export class ImportToolContent extends AbstractToolContent {
 	createView() {
 		const translate = (key) => this._translationService.translate(key);
 
+		const onClear = () => {
+			const inputElement = this._root.querySelector('input');
+			inputElement.value = null;
+		};
 		const onUpload = () => {
 			const inputElement = this._root.querySelector('input');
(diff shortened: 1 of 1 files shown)


src/modules/toolbox/components/importToolContent/ImportToolContent.js
```diff
@@ -27,10 +27,13 @@ export class ImportToolContent extends AbstractToolContent {
 	createView() {
 		const translate = (key) => this._translationService.translate(key);
 
+		const onClear = () => {
+			const inputElement = this._root.querySelector('input');
+			inputElement.value = null;
+		};
```
taulinger, line 33: Can we add a test for that?

taulinger (approved):

taulinger (commented):
Sorry, two more questions 🤓

- why do we not set the value to an empty string (instead of `null`)?
- why do we use `onclick` instead of `onFocus` ?

taulinger (approved):
Thx!

PR #908, opened 2022-04-08
 take care of maps zoom level bounds

### LINKED ISSUE #907: zooming to min or max zoom level causes endless animation loop
when using zoom buttons

taulinger:
- fixes #907 
- optimizes animation duration when zooming the map
- makes it possible to set just the ZOOM or the CENTER query parameter (optionally in combination with the ROTATION parameter)


thiloSchlemmer (approved):
Well done!

PR #920, opened 2022-04-12
fix add/remove of EventListeners

### LINKED ISSUE #919: RemoveLastPoint is triggered multiple times
after switching from 'measure'- to 'draw'-tool multiple times, removing the last point of a polyline or polygon results in a removing of multiple points

thiloSchlemmer:
fixes #919


taulinger (approved):
THX

PR #936, opened 2022-04-14
fix trigger to clear input.value in ImportToolContent

### LINKED ISSUE #904: import file; enable import of same file

thiloSchlemmer:
update event-trigger to input-label
fixes #904

### DIFF
--- a/src/modules/toolbox/components/importToolContent/ImportToolContent.js
+++ b/src/modules/toolbox/components/importToolContent/ImportToolContent.js
@@ -74,13 +74,13 @@ export class ImportToolContent extends AbstractToolContent {
 				</div>
 				<div class="ba-tool-container__content divider" >                						     				
 					<div class="tool-container__buttons">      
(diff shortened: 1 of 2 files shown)


src/modules/toolbox/components/importToolContent/ImportToolContent.js
```diff
@@ -74,13 +74,13 @@ export class ImportToolContent extends AbstractToolContent {
 				</div>
 				<div class="ba-tool-container__content divider" >                						     				
 					<div class="tool-container__buttons">      
-						<label for='fileupload' class="tool-container__button" role="button" tabindex="0" target="_blank" id="share-api" title="upload" > 	                              
+						<label for='fileupload' class="tool-container__button" role="button" tabindex="0" target="_blank" id="share-api" title="upload" @focus=${onClear}>
```
taulinger, line 77: Would suggest to rename the function `onClear` to  `clear[Input]`. "On" would indicate a callback...

taulinger (approved):
Thx

PR #937, opened 2022-04-14
Prevent overlapping MeasurementOverlays

thiloSchlemmer:
Under some circumstances the measurement-overlays overlapping each other...
With this enhancement, the calculation is updated to prevent the overlap.


taulinger (approved):
Thx

PR #962, opened 2022-04-25
olMap: emit notification in case of layer adding failure

### LINKED ISSUE #939: inform user when layer could not be added on map
replaces #790

taulinger:
resolves #939


thiloSchlemmer (approved):
Nice

PR #964, opened 2022-04-26
take real data as input for source type detection

taulinger:
Derive final source type from loaded data. They might not be what they pretend to be...

Test URL: https://www.bergfex.de/downloads/gps/?type=&id=82563&fileType=kml


thiloSchlemmer (dismissed):
Thx!

thiloSchlemmer (approved):
THX

PR #967, opened 2022-04-27
LayerItem: hide inactive buttons 

thiloSchlemmer:
Update the visibility and behavior of copy- and LayerInfo-button in a LayerItem,
whether or not the layer can be copied/has MetaData or not.


src/services/FileStorageService.js
```diff
@@ -82,6 +82,8 @@ export const FileStorageServiceDataTypes = Object.freeze({
 	KML: 'application/vnd.google-earth.kml+xml'
 });
 
+export const TEMP_STORAGE_ID = 'temp_storage_id';
+
```
taulinger, line 86: Would suggest redoing this change: As before the temp. GeoResource ID just needs to be known locally

taulinger (approved):
Thx!

PR #968, opened 2022-04-28
add featureinfo style

alsturm:
objektinffo

### DIFF
--- a/src/modules/featureInfo/components/featureInfoPanel.css
+++ b/src/modules/featureInfo/components/featureInfoPanel.css
@@ -114,6 +114,7 @@ img {
 	color: var(--primary-color) .3;
 }
 
+/* just for grundsteuer release, should be removed later */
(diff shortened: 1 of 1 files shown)


src/modules/featureInfo/components/featureInfoPanel.css
```diff
@@ -112,3 +113,7 @@ img {
 .ba-section:hover .ba-list-item__header .ba-list-item__text {
 	color: var(--primary-color) .3;
 }
+
+.grundsteuer p {
+	padding: 0 0 .8em 0;
+}
```
taulinger, line 119: Let's add a comment, something like "just for grundsteuer release, should be removed later"

alsturm: done

taulinger (approved):

PR #980, opened 2022-05-01
fix tests for units.provider

### LINKED ISSUE #978: units.provider tests fail in Chrome
on local machines (LANG = de_DE.UTF-8)

=> `npm run test --browsers=ChromeHeadless`

thiloSchlemmer:
fixes #978

### DIFF
--- a/src/services/provider/units.provider.js
+++ b/src/services/provider/units.provider.js
@@ -2,12 +2,12 @@ import { $injector } from '../../injection';
 
 const Kilometer_In_Meters = 1000;
 const Squaredkilometer_In_Squaredmeters = 1000000;
-const Locales_Fallback_Lang = 'en';
+const Locales_Fallback = 'en';
 
 const getLocales = () => {
 	const { ConfigService: configService } = $injector.inject('ConfigService');
 
-	return [configService.getValue('DEFAULT_LANG', 'en'), Locales_Fallback_Lang];
+	return [configService.getValue('DEFAULT_LANG', 'en'), Locales_Fallback];
 };
 
 /**
--- a/src/services/provider/units.provider.js
+++ b/src/services/provider/units.provider.js
@@ -4,15 +4,19 @@ const Kilometer_In_Meters = 1000;
 const Squaredkilometer_In_Squaredmeters = 1000000;
 const Locales_Fallback_Lang = 'en';
 
+const getLocales = () => {
+	const { ConfigService: configService } = $injector.inject('ConfigService');
+
+	return [configService.getValue('DEFAULT_LANG', 'en'), Locales_Fallback_Lang];
+};
+
 /**
 * Appends the metric unit of distance to the specified number with bvv defined decimals
 * @param {number} distance the distance value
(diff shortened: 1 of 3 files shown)


taulinger (commented):
Hm, seem to be a workaround, the source of our problem would stay untouched:
All i18n related "translations" should be only dependent on `ConfigService.getValue('DEFAULT_LANG', 'en')`.
Otherwise, a user could get mixed content if the system LANG != env LANG.

taulinger (commented):
Nice! Would just suggest using a fallback locale string.

src/services/provider/units.provider.js
```diff
@@ -7,11 +9,14 @@ const Squaredkilometer_In_Squaredmeters = 1000000;
 * @returns {String} the formatted value
 */
 export const bvvDistanceUnitsProvider = (distance) => {
+	const { ConfigService: configService } = $injector.inject('ConfigService');
+
+	const locale = configService.getValue('DEFAULT_LANG', 'en');
 	const asKilometer = (distanceValue) => {
-		return (Math.round((distanceValue / Kilometer_In_Meters) * 100) / 100).toLocaleString(undefined, { minimumFractionDigits: 1, maximumFractionDigits: 1 }) + ' ' + 'km';
+		return (Math.round((distanceValue / Kilometer_In_Meters) * 100) / 100).toLocaleString(locale, { minimumFractionDigits: 1, maximumFractionDigits: 1 }) + ' ' + 'km';
```
taulinger, line 16: Let's call `toLocaleString` with a fallback od `'en'`.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

src/services/provider/units.provider.js
```diff
@@ -1,17 +1,27 @@
+import { $injector } from '../../injection';
+
 const Kilometer_In_Meters = 1000;
 const Squaredkilometer_In_Squaredmeters = 1000000;
+const Locales_Fallback_Lang = 'en';
```
taulinger, line 5: just `Fallback_Locale` ?

test/service/provider/units.provider.test.js
```diff
@@ -57,5 +66,48 @@ describe('Units provider', () => {
 		expect(bvvAreaUnitsProvider(12345, 0)).toBe('12,345 m&sup2;');
 		expect(bvvAreaUnitsProvider(12345, 2)).toBe('12,345 m&sup2;');
 		expect(bvvAreaUnitsProvider(10000000, 0)).toBe('10.000 km&sup2;');
+		expect(bvvAreaUnitsProvider(1234567891234, 2)).toBe('1,234,567.890 km&sup2;');
+	});
+
+	it('provides formatted distance for bvv-metric system with de-locales', () => {
+		spyOn(configService, 'getValue').and.returnValue('de');
```
taulinger, line 73: We could test also our two arguments (#withArgs)

taulinger (approved):
Thx

PR #986, opened 2022-05-02
fix _setSelection when select is deactivated

### LINKED ISSUE #985: OlDrawHandler throws unhandled exception on deactivate
`TypeError: this._select.getFeatures is not a function`

thiloSchlemmer:
fixes #985


taulinger (approved):
thx

PR #995, opened 2022-05-05
fix createStringLatLong

### LINKED ISSUE #994: bvvStringifyFunction cannot handle 3D coordinates
```
it('stringifies a 4326 coordinate', () => {
	const coord4326 = [11.57245, 48.14021];
	const coord4326_3d = [11.57245, 48.14021, 42];

	expect(bvvStringifyFunction(4326)(coord4326, { digits: 3 })).toBe('48.140, 11.572');
	expect(bvvStringifyFunction(4326)(coord4326_3d, { digits: 3 })).toBe('48.140, 11.572'); <-------fails
});
```

thiloSchlemmer:
Instead of simple reverse the order of assumed 2D coordinate-array only
switching the XY-part of the coordinate-array to left the righter
parts (Z,...) untouched.

fixes #994

### DIFF
--- a/src/services/provider/stringifyCoords.provider.js
+++ b/src/services/provider/stringifyCoords.provider.js
@@ -36,11 +36,12 @@ export const bvvStringifyFunction = (srid, options = {}) => {
 
 /**
  * A function that returns a function specific for the representation of geographic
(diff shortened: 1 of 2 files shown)


taulinger (approved):
Nice, thx!

PR #1007, opened 2022-05-09
fit to layer extent

taulinger:
Add the possibility to fit on a specific vector layer on the map
A fit request is resolved by the map implementation.
Fit requests for other layer types may be implemented in the future.

### DIFF
--- a/src/modules/olMap/components/OlMap.js
+++ b/src/modules/olMap/components/OlMap.js
@@ -188,19 +188,7 @@ export class OlMap extends MvuElement {
 
 		//register particular obeservers on our Model
 		//handle fitRequest
-		this.observeModel(['fitRequest', 'fitLayerRequest'], (eventLike) => {
-			const onAfterFit = () => {
-				this._viewSyncBlocked = false;
-				this._syncStore();
-			};
-			const extent = eventLike.payload.id ? getLayerById(this._map, eventLike.payload.id).getSource()?.getExtent() : eventLike.payload.extent;
-
-			if (extent) {
-				this._viewSyncBlocked = true;
-				const maxZoom = eventLike.payload.options.maxZoom || this._view.getMaxZoom();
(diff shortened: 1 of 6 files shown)


src/modules/olMap/components/OlMap.js
```diff
@@ -187,19 +188,23 @@ export class OlMap extends MvuElement {
...
 			};
-			const maxZoom = fitRequest.payload.options.maxZoom || this._view.getMaxZoom();
-			this._view.fit(fitRequest.payload.extent, { maxZoom: maxZoom, callback: onAfterFit });
+			const extent = eventLike.payload.id ? getLayerById(this._map, eventLike.payload.id).getSource()?.getExtent() : eventLike.payload.extent;
+
+			if (extent) {
+				this._viewSyncBlocked = true;
+				const maxZoom = eventLike.payload.options.maxZoom || this._view.getMaxZoom();
+				this._view.fit(extent, { maxZoom: maxZoom, callback: onAfterFit }); // Todo: fit map with padding parameter by calling mapService#calculatePadding
+			}
```
thiloSchlemmer, line 202: Can we extract this to a helper-method, to be consistent to the code unterneath?

thiloSchlemmer (approved):
Nice! One small hint inside

src/modules/olMap/components/OlMap.js
taulinger: >to be consistent to the code unterneath?

 Sorry, didn't get the point, can you explain it more in detail, please?

thiloSchlemmer: sure...

Line 205 and 207 uses only simple calls to private functions,  the code in line 192 - 202 can be moved to a simple method, too. In the result, the reader gets a consistent overview over all observed properties and the related behavior.

thiloSchlemmer (approved):
Nice, THX!

PR #1011, opened 2022-05-12
MapService static padding

thiloSchlemmer:
Adding a simple strategy to calculate an optional map padding.
This strategy use fixed css property values to calculate the area where the map is not covered by other components.

The main assumption is, that the main menue is the only component which will be in conflict with any map content. So only the width of the main menue is taken into account for the padding.


taulinger:
Nice!
Let's move the logical part to its own provider.

thiloSchlemmer:
obsolete  by #1079 and #1084

PR #1028, opened 2022-05-19
remove eslint-plugin-standard dependency

taulinger:
according to https://github.com/standard/standard/issues/1316


thiloSchlemmer (approved):
Thx!

PR #1029, opened 2022-05-19
abort by key

thiloSchlemmer:
Abort Drawing or measurement by key-press.


src/modules/commons/KeyActionMapper.js
```diff
@@ -0,0 +1,81 @@
+import { isString } from '../../utils/checks';
+const No_Op = () => { };
+
+/**
+ * Maps actions to specific key-events
+ * @class
+ * @author thiloSchlemmer
+ */
+export class KeyActionMapper {
```
taulinger, line 9: Nice, would just suggest moving this class to `src/utils`

taulinger (commented):
Hmm, I think this PR does not restore exactly the same behavior as in v3. In v3 pressing `ESC` fully deactivates the drawing/measurement tool, however here it just stops the current action.

thiloSchlemmer:
> Hmm, I think this PR does not restore exactly the same behavior as in v3. In v3 pressing `ESC` fully deactivates the drawing/measurement tool, however here it just stops the current action.

Yes, you're right. The main reason for the discrepance is, that it is not necessary anymore to close the corresponding tool (measurement/drawing), to make something else, as to draw or to measure in the map.

The main requirements for the Abort-By-Key are:

- Abort the current drawing to make a new and better drawing and delete the current one.
- Abort the current drawing to make something else.

In v3 it is necessary to completly close the "Draw"-modus to make something different. If you want to make a new drawing, you have to select again the "Draw"-modus.

In v4 we does not have a ui/ux-element for a specific modus, which we must close, to give the user the opportunity to make something else in the UI.

src/modules/commons/KeyActionMapper.js
taulinger: -> `src/utils` not `src/modules/utils` 🤓

src/modules/olMap/handler/draw/OlDrawHandler.js
```diff
@@ -76,7 +77,9 @@ export class OlDrawHandler extends OlLayerHandler {
 		this._storedContent = null;
 		this._sketchHandler = new OlSketchHandler();
 		this._mapListeners = [];
-		this._keyUpListener = (e) => this._removeLast(e);
+		this._keyActionMapper = new KeyActionMapper(document)
+			.addForKeyUp('Delete', () => this._remove())
+			.addForKeyUp('Escape', () => this._reset());
```
taulinger, line 82: statement not covered

taulinger (approved):
Thx!

PR #1035, opened 2022-05-24
coerce sanitizeHtmlProvider by additional options

thiloSchlemmer:
coercing by additional blacklist of forbidden/ unsafe HTML-Tags and
filtering of SVG- and mathML-Tags


taulinger (approved):
Thx

PR #1036, opened 2022-05-24
update sanitizer with blacklist & html profile

thiloSchlemmer:
update sanitizeHtmlProvider with additional blacklist of forbidden/ unsafe HTML-Tags and
filtering of SVG- and mathML-Tags


taulinger (approved):
Thx

PR #1048, opened 2022-06-03
connect wms import related services and show search results within UI

taulinger:
- [x] import non-restricted WMS
- [x] import restricted WMS
- [x]  handle non-normalized URLs (http://foo.bar/foo.cgi?%27;)


thiloSchlemmer (approved):
Nicely done!

PR #1052, opened 2022-06-08
add toggle for password visibility

thiloSchlemmer:
Add a toggle to make the password characters for the user readable or not. This gives the user the possibility to ensure the validity of the credential and prevents typos.


src/modules/auth/components/PasswordCredentialPanel.js
```diff
@@ -146,6 +146,17 @@ export class PasswordCredentialPanel extends MvuElement {
...
+			const togglePassword = this.shadowRoot.querySelector('#toggle_password');
+			const password = this.shadowRoot.querySelector('#credential_password');
+
+			const type = password.getAttribute('type') === 'password' ? 'text' : 'password';
+			password.setAttribute('type', type);
+
+			// toggle the icon
+			togglePassword.classList.toggle('eye');
+		};
+
```
taulinger, line 159: Works nicely, but shouldn't we use our MVU approach also here?

taulinger (commented):
line 106 has an uncovered function

taulinger (approved):
Very nice!

PR #1073, opened 2022-06-23
recognize wms-url in DndImportPanel and use setQuery

thiloSchlemmer:
Adding the option to drag&drop a wms url.
Due to the fact, that the user must select a specific layer, the georecource is not added to the map immediatly, but displayed as SearchResult. This is already implemented and now used by setQuery in search.action.


taulinger (approved):
Nice (sorry for the previous confusion)

PR #1079, opened 2022-06-28
calculate clean rectangle

thiloSchlemmer:
calculating a clean rectangle, where a base-element is polluted by an array of dirty elements.
This rectangle can be used to define a padding for a map.view 

- [ ] calculation of "CleanRectangle"
- [ ] find better wording


src/services/MapService.js
```diff
@@ -127,4 +130,15 @@ export class MapService {
 		const element = document.querySelector('ba-footer')?.shadowRoot.querySelector('.scale');
 		return element ?? null;
 	}
+
+	/**
+	 * Returns a visible rectangle, where the map element is not overlapped by any other relevant element
```
taulinger, line 135: <s>visible</s>

src/services/MapService.js
```diff
@@ -127,4 +130,15 @@ export class MapService {
 		const element = document.querySelector('ba-footer')?.shadowRoot.querySelector('.scale');
 		return element ?? null;
 	}
+
+	/**
+	 * Returns a viewport rectangle, where the map element is not overlapped by any other relevant element
```
taulinger, line 135: Returns a DOMRect instance describing the visible part of the map (which means the part of the part that is not overlapped by any other UI element)

src/services/MapService.js
```diff
@@ -127,4 +130,15 @@ export class MapService {
...
+	 * @returns {DOMRect}
+	 */
+	getVisibleViewport() {
+		const viewportElement = document.body;
+		const overlappingElements = findAllByAttribute(document, 'data-register-for-viewport-calc');
+
+		return calculateVisibleViewport(viewportElement, overlappingElements);
```
taulinger, line 142: just
```
const overlappingElements = findAllByAttribute(document, 'data-register-for-viewport-calc');
return calculateVisibleViewport(document.body, overlappingElements);

``` 
?

test/service/MapService.test.js
```diff
@@ -187,5 +187,32 @@ describe('MapService', () => {
...
+
+	describe('getVisibleViewport', () => {
+		it('returns a visible rectangle', () => {
+			document.body.innerHTML = '<div></div>'
+									+ '<div id="overlapping1" data-register-for-viewport-calc></div>'
+									+ '<div id="non-overlapping"></div>'
+									+ '<div id="overlapping2" data-register-for-viewport-calc></div>'
+									+ '<div></div>'
+									+ '<div class="class foo"></div>'
+									+ '<div></div>';
```
taulinger, line 199: do we really need them?

taulinger (commented):
NIce, would suggest only some minor optimizations

taulinger (approved):
Thx!

PR #1097, opened 2022-07-06
add a query parameter for handling a query term

taulinger:
`http://localhost:8082/?q=something` 
leads to:
image


thiloSchlemmer (approved):
NIce

PR #1098, opened 2022-07-07
update visibility of zoomToExtent-button

thiloSchlemmer:
The zoomToExtent-Button in the LayerItem will be not visible, when the
georecource of a layer is not a VectorGeoResource.


taulinger (approved):
Thx!

PR #1099, opened 2022-07-07
add .env file

taulinger:
Contains the `SOFTWARE_INFO` property which causes the showcase button to be displayed when running the front-end only.


thiloSchlemmer (approved):

taulinger:
Hmm, does make problems when `.env` already exists (overwrites it without warning). Closing it.

PR #1106, opened 2022-07-11
update current state of window history

taulinger:
Permanently update the URL of the current history to reflect the (public) state of the app.


thiloSchlemmer (approved):
Nice

PR #1110, opened 2022-07-12
extend attribution domain object

taulinger:
by allowing multiple copyright entries
and use an OpenData WMTS for fallback georesources:

grafik


thiloSchlemmer (approved):
Good job!

PR #1113, opened 2022-07-14
AutocompleteSearch refactor to increase testcoverage

thiloSchlemmer:
The AutocompleteSearch component has untested (untestable?) code. This PR have the goal to refactor the code to make the behavior of 'update the current focus'(?) testable.

- [x] isolate the behavior
- [x] test the behavior
- [ ] find better names for the testcase description
- [ ] find a better name for the refactored method (_syncFocusWith)


taulinger:
This component was one of the first we wrote in order to try out some common and expected use cases based on our new approach. Currently we have no usage for this component.
So, I'm wondering if it's worth the effort...

thiloSchlemmer:
closed in favor to #1114

PR #1118, opened 2022-07-15
fix StyleOptions of selected feature

### LINKED ISSUE #1104: style size dropdown won't be updated
After re-selecting an icon/text the style size dropdown value does not necessarily show the correct value. 

grafik

affects:

- 4.0
- 4.1

thiloSchlemmer:
Add scale information of selected features to the changed styleOption in store, to display the correct scale-value in the DrawToolContent

fixes #1104


taulinger (approved):
Works well now!

PR #1125, opened 2022-07-19
add ExportMfpToolContent

thiloSchlemmer:
A component to control the settings for a export with MapFishPrint by the user.


taulinger:
asset `src/modules/toolbox/components/exportMfpToolContent/assets/download.svg` seems to be unnecessary

taulinger (commented):
Nice proceedings!

Hmm, I'm wondering if we could slightly simplify it by reworking the components model. Maybe we should hold the
- capabilities
- name (in the meaning of an id, will also be used for i18n later) -> references selected page layout
- scale -> selected scale
After loading the capabilities we would set "name" and "scale" initially...
But let's discuss this tomorrow...

test/modules/toolbox/components/toolContainer/ToolContainer.test.js
```diff
@@ -118,6 +120,12 @@ describe('ToolContainer', () => {
 					case ToolId.SHARING:
 						expect(content.innerHTML.toString().includes(ShareToolContent.tag)).toBeTrue();
 						break;
+					case ToolId.IMPORT:
+						expect(content.innerHTML.toString().includes(ImportToolContent.tag)).toBeTrue();
+						break;
```
taulinger, line 125: Nice!

taulinger (approved):
Fine!

PR #1143, opened 2022-07-25
determine preferred getMap media type

taulinger:
We should not just take the first media type for GetMap requests from the GetCapabilities document, 
but determine the best suitable one.


thiloSchlemmer (approved):
Well done!

PR #1153, opened 2022-07-28
disable intermediary zoom levels

### LINKED ISSUE #1090: disable intermediary zoom levels for the map (optionally)
to improve the quality of map tiles (sharpness). 

Affect the smoothness of zooming on touch devices, so maybe we should let the user decide which mode to use and place an option for that under `More/Settings`

taulinger:
Partly resolves #1090. Making it configurable for the user will be done later...


thiloSchlemmer (approved):
THX

PR #1185, opened 2022-08-29
use ol-maplibre-layer

taulinger:
for visualization of Vector Tiles.

grafik

Some notes:
- ol-maplibre-layer just provides an ol layer that renders a MapLibre GL JS Map (https://github.com/geoblocks/ol-maplibre-layer/blob/master/src/MapLibreLayer.js)
- bundle size will increase from 1.65 MiB to 2.36 MiB
- FF currently throws a WebGL error when running in headless mode (https://bugzilla.mozilla.org/show_bug.cgi?id=1375585#c27), so we first check if it does make sense to perform the test, otherwise, we skip them


thiloSchlemmer (approved):
Well done!

PR #1196, opened 2022-09-15
PdfExport: connect client with backend (I)

taulinger:
For the present just  `MfpService#getCapabilities()` has been implemented, the other methods will be done in a second PR and are still mocked.


thiloSchlemmer (approved):
Thx!

PR #1205, opened 2022-09-21
PdfExport: connect client with backend (II) 

taulinger:
- [x] Implement all remaining methods of `MfpService`
- [x] cleanup (extract iface)


thiloSchlemmer (approved):
Nice

PR #1216, opened 2022-10-07
add ADV WMTS compatible TileGrid

taulinger:
- extends `WMTSGeoResource` for optional property `tileGridId`
- extends 'GeoResource` for optional property `exportable`
- adds an ADV WMTS compatible TileGrid


thiloSchlemmer (approved):
Thx!

PR #1230, opened 2022-10-14
fitToExtent: handle the case when the layer's source is null

taulinger:


src/modules/olMap/utils/olMapUtils.js
```diff
@@ -1,3 +1,5 @@
+import { OlMap } from '../components/OlMap';
```
thiloSchlemmer, line 1: can be removed...
although referenced for jsdoc, `npm run doc` does not throw an error, if removed

taulinger: Oops, thx

taulinger: Hmm, seems to work locally and here on CI

thiloSchlemmer (dismissed):
Thx

thiloSchlemmer (approved):
Thx

PR #1232, opened 2022-10-14
improve projection registration

taulinger:
Add the possibility to detect all available (means registered) projections,
which is realized by the new `Proj4JsService` class.


thiloSchlemmer (approved):
Cool... Well done!

PR #1258, opened 2022-11-02
fix content of thumbnailLabel

thiloSchlemmer:
Fixes a misbehavior where the label content of the thumbnail-style was not updated, after the mfp settings are changed.
The thumbnail-style is used to display the currently selected mfp-settings and extent as preview for the export.


taulinger (approved):
Thx!

PR #1272, opened 2022-11-04
fix OlMfpHandler calculation of optimalScale

### LINKED ISSUE #1262: Export2Pdf: print rectangle not always (fully) visible
There seems to be a problem with the calculation of the best-fitting scale when the device screen has a particular resolution e.g 1900x1100px. 

export2Pdf_2022-11-02_18-14-26.

⬆️ Where is the rectangle 🙃?

thiloSchlemmer:
fixes #1262


taulinger (changes requested):
> Due to updated version of firefox (105) in our playwright setup, using now `Array.findLast()` instead of a private implementation in `OlMfpHandler._getOptimalScale()`.

Hmm, this might be a little bit early, FF 115.0 ESR, which is the first FF ESR release supporting `findLast`, will be released [06/23](https://wiki.mozilla.org/Release_Management/Calendar). So let's keep our own implementation for a while...

The actual fix works nicely for me!

taulinger (approved):
Thx!

PR #1273, opened 2022-11-04
Fix empty data owner

### LINKED ISSUE #1271: Export2Pdf: attribute 'dataOwner' must not be null
https://github.com/ldbv-by/bav4-nomigration/blob/8f8f93055871e7c02b357902394055d008b06191/src/modules/olMap/services/Mfp3Encoder.js#L145-L146

Like in L146 let's give `dataOwner` an empty String as fallback.

thiloSchlemmer:
fixes #1271


taulinger (approved):
Looks good!

PR #1302, opened 2022-11-17
Export2Pdf: add a default parameter to encoded WMS

thiloSchlemmer:
adding transparent:true to encoded MFP  specs of a WMS layer, by adding the customParams property


taulinger (approved):
Thx!

PR #1316, opened 2022-11-22
remove usage of HTML Entities

thiloSchlemmer:
remove usage of `&sup2;` to prevent unnecessary encode- and decode-calls


taulinger (approved):
Nice!

PR #1329, opened 2022-11-25
refactor global i18n

taulinger:
Merge all i18n providers of the plugin directory to one global provider.
This global provider is allowed to be consumed both by plugins and services.


thiloSchlemmer (approved):
Thx

PR #1331, opened 2022-11-29
add attribution for user-imported GeoResources

taulinger:
Add an attribution for the following cases:

- User creates a drawing/measurement 
- User loads a drawing/measurement data set (by query parameter)
- User imports a vector data set locally
- User imports a vector data set by URL
- User imports a WMS

In cases of loading the data by URL, we provide the base URL as attribution, in cases of importing/creating the data set locally we add a hint (Dataset and/or style provided by third party, Mit Darstellung durch den Anwender)


thiloSchlemmer (approved):
Well done!

PR #1347, opened 2022-12-08
Export2Pdf: optimize mask performance

thiloSchlemmer:
Reducing calls of time consuming calculations and change Maskrenderer from usage of [CanvasImmediateRenderer](https://openlayers.org/en/latest/apidoc/module-ol_render_canvas_Immediate-CanvasImmediateRenderer.html) to simple calls of canvas context.


taulinger (changes requested):
Very smooth now  😍
but has obviously a bug when manually rotating the map
Steps:
- disable autorotaion
- rotate the map
- enable autorotationn
grafik

taulinger (changes requested):
Hmm, there seems something not to be correct in sync yet when autorotation is disabled:

grafik

thiloSchlemmer:
> Hmm, there seems something not to be correct in sync yet when autorotation is disabled:
> 
> grafik

This weird representation results from the calculation of the extents (mfp-feature, valid area without distortion) and not from the actual polygons.

Due to the possebilities in using feature styles,  i see  the following solutions , at this point:

- leave it
- remove the styling of the areaOfDistortion without replacing
- replace it with a correct calculation of a difference-polygon (boolean operation)

**Difference-polygon** requires more advanced calculations, where i would rely on packages like [polygon-clipping](https://github.com/mfogel/polygon-clipping) or [turfjs](https://github.com/Turfjs/turf).
Branch [tmp_mfpStyleUtils_calc-polygonDifference](https://github.com/ldbv-by/bav4-nomigration/tree/tmp_mfpStyleUtils_calc-polygonDifference) implements a solution with usage of 'polygon-clipping'.

taulinger:
Simple enabling the tool with connected ba-backend (real Mfp capabilities) leads to 
grafik

😭

taulinger:
Following our recent  decision, we should close this PR (but keep up the branch)

taulinger (commented):
According to our recent decision, we should close this PR (but keep up the branch)

PR #1351, opened 2022-12-09
OlMap: prevent double calls of map actions

taulinger:
Prevent double calls of map actions (movestart, moving, moveend).


thiloSchlemmer (approved):
Thx

thiloSchlemmer (approved):
Thx!

PR #1438, opened 2023-02-03
prevent unnecessary calls of History API

taulinger:
Avoid throwing an exception due to too many calls to History API within a short timeframe.

### DIFF
--- a/src/plugins/HistoryStatePlugin.js
+++ b/src/plugins/HistoryStatePlugin.js
@@ -38,7 +38,7 @@ export class HistoryStatePlugin extends BaPlugin {
 	_updateHistory() {
 		const encodedState = this._shareService.encodeState();
 		if (this._currentEncodedState !== encodedState) {
-			this._environmentService.getWindow().history.replaceState(null, '', this._shareService.encodeState());
+			this._environmentService.getWindow().history.replaceState(null, '', encodedState);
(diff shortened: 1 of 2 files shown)


src/plugins/HistoryStatePlugin.js
```diff
@@ -35,6 +36,10 @@ export class HistoryStatePlugin extends BaPlugin {
 	}
 
 	_updateHistory() {
-		this._environmentService.getWindow().history.replaceState(null, '', this._shareService.encodeState());
+		const encodedState = this._shareService.encodeState();
+		if (this._currentEncodedState !== encodedState) {
+			this._environmentService.getWindow().history.replaceState(null, '', this._shareService.encodeState());
```
thiloSchlemmer, line 41: iis a additional call of `this._shareService.encodeState()` needed?  
Otherwise `this._environmentService.getWindow().history.replaceState(null, '', [encodedState]);` would be more readable for me.

thiloSchlemmer (approved):
Nice!

Small suggestion inline

src/plugins/HistoryStatePlugin.js
taulinger: Thx

thiloSchlemmer (approved):
Thx!

PR #1445, opened 2023-02-07
add: icon to button component

alsturm:


taulinger (changes requested):
Does not work in FF
image

src/modules/commons/components/button/Button.js
```diff
@@ -109,6 +131,17 @@ export class Button extends MvuElement {
 		return this.getModel().label;
 	}
 
+	/**
+	 * @property {string} icon='default_svg_icon' - Data-URI of Base64 encoded SVG
```
taulinger, line 135: The default value for the icon is `null` -> `icon=null`

taulinger (commented):
Nice, would just suggest fixing the docs

taulinger (approved):
Nice enhancement!

PR #1479, opened 2023-02-17
resize chart on update

### LINKED ISSUE #1468: ElevationProfile does not always update properly
... when changing selected geometry

grafik

taulinger:
fixes #1468

(see also https://www.chartjs.org/docs/latest/developers/api.html)


taulinger:
Does not work reliably.

PR #1480, opened 2023-02-18
ElevationService: return mocked data in standalone mode

taulinger:
Elevation profile shows mocked data now when app is in standalone mode.


thiloSchlemmer (approved):
Nice!

PR #1482, opened 2023-02-20
BottomSheet: unsubscribe the observer 

### LINKED ISSUE #1452: BottomSheet: observer won't be unsubscribed when component disconnects

thiloSchlemmer:
fixes #1452

Move the subscription to store changes into the constructor phase and complete the lifecycle with the unsubscribe call in the disconnect-callback.


taulinger (approved):
Thx!

PR #1501, opened 2023-02-27
ElevationProfile: slope in tooltip

norbertK:
implements task BAY-867: Darstellung der Steigungswerte im Tooltip


test/modules/elevationProfile/components/ElevationProfile.test.js
```diff
@@ -125,15 +127,17 @@ describe('ElevationProfile', () => {
 		attrs: [
 			{
 				id: 'slope',
+				prefix: '~',
```
taulinger, line 130: Let's omit the prefix property here, which would reflect our real use case

taulinger (commented):
Works for me. thx

taulinger (approved):
Thx

PR #1509, opened 2023-03-02
ElevationProfile: use UnitsService for distances

norbertK:
implements BAY-870


taulinger (commented):
1. The unit of the distance seems to be wrong yet.
grafik

2. If the distance is less than 1000m the tooltip displays "NaN"

grafik

taulinger (approved):
Thx

PR #1518, opened 2023-03-03
add: button to activate the map

alsturm:


taulinger (commented):
Let's display the activation button not until the user hovers over or touches the map (which means the same behavior for touch and non-touch environments)

taulinger (commented):
Please resolve existing conflicts first...

taulinger (approved):
Finally!

PR #1533, opened 2023-03-08
ShareService: ignore existing pathname

taulinger:
`encodeState()` will ignore an existing pathname from the current location now. If needed it always has to be requested as a path parameter,
e.g.: `shareService.encodeState({}, [PathParameters.EMBED]);`


alsturm (dismissed):

thiloSchlemmer (approved):
Thx!

PR #1566, opened 2023-03-17
optimize elevationProfile slice-of-state

taulinger:
Remove possible side effect when closing the profile.


thiloSchlemmer (approved):
Thx!

PR #1622, opened 2023-04-11
fix Showcase formatting

taulinger:
Background: `input` elements must be self-closing otherwise Prettier won't format template strings


norbertK (approved):
ok

Issue #1614, opened 2023-04-05
Remove confusion of the MapService's methods names


taulinger:
Make a clear difference between geodesic-related and projected-related methods.

taulinger:
Resolved by #1626

PR #1626, opened 2023-04-11
rework displaying of coordinates

taulinger:
Map coordinates are now basically displayed  in the UI  as WGS84, UTM, or MGRS coordinates
Within a defined extent coordinates can be also displayed in local projected reference systems.

(this PR resolves also #1614)


thiloSchlemmer (dismissed):
Looks good to me...

especially the formulation with local-projected and global coordinates expresses our intention better

thiloSchlemmer (approved):
Well done

PR #1673, opened 2023-04-27
Optimize map feedback style

alsturm:


src/modules/iframe/components/tools/DrawTool.js
```diff
@@ -107,30 +108,39 @@ export class DrawTool extends MvuElement {
...
+			return active ? 'draw-tool__enable' : 'draw-tool__disable';
+		};
 
 		return html`
 			<style>
 				${css}
 			</style>
 			<div class="draw-tool">
-				<div class="draw-tool__content">
+				<div class="draw-tool__content ${getActiveClass()}">
```
taulinger, line 122: Should we add a test for the presence of this class?

taulinger (commented):
Well done!

src/modules/iframe/components/tools/DrawTool.js
alsturm: done

taulinger (approved):

PR #1676, opened 2023-05-02
MapFeedback: remove Feedback components from ShowCase

thiloSchlemmer:
After it has been successfully placed in the menu, it is no longer needed in the showcase.


taulinger (approved):
Thx

PR #1697, opened 2023-05-08
change XytGeoResource

taulinger:
- add new property `urls` which could either be a `string` or a `string[]`
- remove property `url`


thiloSchlemmer (approved):
Thx

PR #1702, opened 2023-05-10
IframeGeometryIdPlugin: fix updateAttribute to emptyString

thiloSchlemmer:
update the behavior of the method '_updateAttribute' to empty string instead of null, du to the behavior of the setAttribute-method of HtmlElements, which
causes an implicit stringify of null-values.
An empty string ('') can be easier evaluated then 'null'.


taulinger (approved):
Thx!

PR #1704, opened 2023-05-11
Add style for clsuter layers

thiloSchlemmer:
grafik

- [x] default style function
- [ ] specific style function based on georesource id


taulinger (approved):
Nice!
(just added some Tests for the `VectorLayerService`)

PR #1724, opened 2023-05-22
add test for firing the correct event

taulinger:


test/modules/iconSelect/components/IconSelect.test.js
```diff
@@ -161,6 +161,29 @@ describe('IconSelect', () => {
...
+
+			element.click();
+			const selectableIcon = element.shadowRoot.querySelector('#svg_foo');
+			selectableIcon.click();
+
+			expect(spy).toHaveBeenCalledOnceWith(
+				new CustomEvent('toggle', {
+					detail: { selected: jasmine.any(IconResult) }
+				})
+			);
```
thiloSchlemmer, line 184: I just stumbled over this naming issue and wonder why 
it doesn't matter if we test with 
`new CustomEvent('toggle', {detail: { selected: jasmine.any(IconResult) }})`
or 
`new CustomEvent('select', {detail: { selected: jasmine.any(IconResult) }})`
or 
`new CustomEvent('fooBar', {detail: { selected: jasmine.any(IconResult) }})`

All this variants fullfil the expectation.

We should think about a more precise expectation like 
`expect(spy).toHaveBeenCalledOnceWith(jasmine.objectContaining({ type: 'select', detail: { selected: jasmine.any(IconResult) } }));`
which fails, if the type is not 'select'.

thiloSchlemmer (commented):
See inline comment

thiloSchlemmer (approved):
Thx!

PR #1801, opened 2023-06-15
make highlight features selectable

taulinger:
Features of hidden layers are selectable if they contain a `name` property.

grafik


thiloSchlemmer (approved):
Thx

PR #1803, opened 2023-06-16
add maplibre-gl as direct dependency

taulinger:
`maplibre-gl` comes as a peer dependency from `ol-maplibre-layer`. In order to use its latest version we add it as a direct dependency.


thiloSchlemmer (approved):
Thx!

PR #1821, opened 2023-06-23
MapFeedbackPanel: add center property 

thiloSchlemmer:


src/modules/feedback/components/mapFeedback/MapFeedbackPanel.js
```diff
@@ -306,14 +316,16 @@ export class MapFeedbackPanel extends MvuElement {
...
 	set onSubmit(callback) {
 		this._onSubmit = callback;
 	}
 
+	set center(value) {
+		if (isCoordinate(value)) {
+			this.signal(Update_Center, value);
+		}
+	}
+
```
taulinger, line 328: Would suggest removing this check. (In `set onSubmit` we also do not check if it is a function)

taulinger (commented):
Nice 👍

taulinger (approved):
👍

PR #1828, opened 2023-06-26
improve docs for TopicService and CatalogService

### LINKED ISSUE #1807: CatalogService: inconsistent use of arguments
In [CatalogService.js](https://github.com/ldbv-by/bav4/blame/master/src/modules/topics/services/CatalogService.js) and [catalog.provider.js](https://github.com/ldbv-by/bav4/blob/master/src/modules/topics/services/provider/catalog.provider.js) occurs different usages of the same interface-function.

- https://github.com/ldbv-by/bav4/blob/f742ddbabedec9e6bc839a9058018fefbe568c66/src/modules/topics/services/provider/catalog.provider.js#L36
- https://github.com/ldbv-by/bav4/blob/f742ddbabedec9e6bc839a9058018fefbe568c66/src/modules/topics/services/CatalogService.js#L46
- https://github.com/ldbv-by/bav4/blob/f742ddbabedec9e6bc839a9058018fefbe568c66/src/modules/topics/services/CatalogService.js#L56

Despite the fact that the parameter is not used in the default implementation, further implementations from other providers might need to use the parameter.

I suggest changing the usages to a unified signature: `loadFallbackCatalog(topiId)` and adding an ignore hint in the default implementation of `catalog.provider.js`.

taulinger:
resolves #1807

### DIFF
--- a/src/modules/topics/services/CatalogService.js
+++ b/src/modules/topics/services/CatalogService.js
@@ -1,14 +1,33 @@
 /**
  * @module modules/topics/services/CatalogService
  */
+import {
+	FALLBACK_GEORESOURCE_ID_0,
+	FALLBACK_GEORESOURCE_ID_1,
+	FALLBACK_GEORESOURCE_ID_2,
+	FALLBACK_GEORESOURCE_ID_3
+} from '../../../services/GeoResourceService';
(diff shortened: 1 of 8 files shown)


thiloSchlemmer (approved):
Nice!

PR #1836, opened 2023-06-28
bottomSheet.css: remove text color

### LINKED ISSUE #1826: ContextMenu: hidden text
The text content of the context menu is not readable in the mobile version with the light theme.
image
Screenshot_20230623-192133_Chrome

thiloSchlemmer:
due to the inheritance of some CSS properties(e.g. color) into the child component, the color property is not required or may even lead to unwanted display problems

fixes #1826


taulinger (approved):
Test is missing 😜

PR #1841, opened 2023-06-28
 drawing of symbol not possible after style color is changed

evgeniTUM:
bug workflow:
1.  open draw dialog
2.  select "symbol"
3.  open "style" area 
4.  choose another color

result:
-     no drawing of symbol is possible any more
-     there is no icon selected inside "style" area

This pull requests changes the inital value of the marker to "marker.png" as the matcher function of the IconResult class only matches if the URL ends with "/${id}.png"

Screenshot from 2023-06-28 18-08-45

### DIFF
--- a/src/services/IconService.js
+++ b/src/services/IconService.js
@@ -35,7 +35,7 @@ export class IconService {
 			return (color) => {
 				try {
 					const url = configService.getValueAsPath('BACKEND_URL') + 'icons';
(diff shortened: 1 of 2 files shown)


taulinger:
@thiloSchlemmer Can you review this PR? Regarding this PR please also have a look at #1842 .

thiloSchlemmer (approved):
thx

PR #1847, opened 2023-06-30
TestUtils: allow setting of initial properties for a component

taulinger:
When a component has one or more public properties a test can set them initially now without using a defined setter.
Example:

```
html`<ba-mvu-togglefeedbackpanel .center=${center}`
```

was tested by something like
```
const element = await TestUtils.render(ToggleFeedbackPanel.tag);
element.center =  : [21, 42] 
```
and will become now
```
const element = await TestUtils.render(ToggleFeedbackPanel.tag, { center: [21, 42] });
```
which reproduces the "real" behavior of the component, because the initial properties are **now available when the component is mounted into the DOM**

### DIFF
--- a/src/modules/MvuElement.js
+++ b/src/modules/MvuElement.js
@@ -127,6 +127,10 @@ export class MvuElement extends HTMLElement {
 	 * @private
 	 */
 	connectedCallback() {
+		// see TestUtils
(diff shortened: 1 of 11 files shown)


thiloSchlemmer (approved):
Well done!

PR #1848, opened 2023-06-30
add lazy loading/dynamic import of component

taulinger:
In this PR the ElevationProfile component and its dependencies (incl. Chart.js) are loaded dynamically on demand by using webpack's dynamic import feature.

The initial build size decreases by almost 10% from 2.97 MiB to  2.77 MiB.

This feature will get much more interesting when we are trying to integrate something like Cesium.js.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,10 @@ All redux related files like reducers and actions.
 
 Contains global utilities.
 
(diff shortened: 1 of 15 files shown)


thiloSchlemmer (approved):
Well done!

PR #1865, opened 2023-07-07
#bugfix sharing of link while an external wms is enabled does not work

evgeniTUM:
It is not possible to share a link whenever an external wms is active
1. enable an external WMS layer
2. share link (copy)
3. Result: The shortening service cannot create a short url for this link, because the "layer" parameter is not URI encoded

Proposed fix: URI encode/decode the "layer" query parameter


taulinger (dismissed):

taulinger (approved):
Thx!

PR #1878, opened 2023-07-11
always clear drawing listeners on deactivate

### LINKED ISSUE #1877: closing measurement tool may cause error
1. start drawing  a measurement by creating the first point
2. close the tool without creating a second point and 
3. zoom the map by the mouse wheel

evgeniTUM:
Resolves #1877

proposed fix:
unsubscribe and clear drawing listeners


taulinger:
@thiloSchlemmer If this solution is accepted maybe we should think about adding a test for this behavior

thiloSchlemmer (approved):
Thx

PR #1880, opened 2023-07-12
remove help module

taulinger:
The FirstSteps component is obsolete now since we're having the Chips component...


thiloSchlemmer (approved):
Thx

PR #1888, opened 2023-07-13
Mfp3Encoder: add handling of groupLayer opacity

thiloSchlemmer:


test/modules/olMap/formats/Mfp3Encoder.test.js
```diff
@@ -740,6 +808,34 @@ describe('BvvMfp3Encoder', () => {
 				customParams: { transparent: true }
 			});
 		});
+		it("resolves wms layer with groupOpacity to a mfp 'wms' spec", () => {
```
taulinger, line 811: do we need a corresponding test for vector type?
```
it("resolves vector layer with groupOpacity to a mfp 'wms' spec"
```

taulinger (approved):
THX

PR #1959, opened 2023-09-12
fix import of vector data from URL

taulinger:
see Ticket#4598227

- [x] increase timeout
- [x] always use URL as ID of the GeoResource


thiloSchlemmer (approved):
Thx

PR #1966, opened 2023-09-14
add:  selectable and draggable behavior

alsturm:


taulinger:
@alsturm  Does the property "user-drag"  really exist?

src/modules/layerManager/components/LayerManager.js
```diff
@@ -178,9 +178,9 @@ export class LayerManager extends MvuElement {
 			<style>
 				${css}
 			</style>
-			<div class="layermanager overflow-container">
+			<div class="layermanager overflow-container draggable">
```
thiloSchlemmer, line 181: should not be draggable, i guess...
this makes the hole container draggable, which includes the action buttons (['expand all'], ['remove all'])

src/modules/layerManager/components/LayerManager.js
```diff
@@ -178,9 +178,9 @@ export class LayerManager extends MvuElement {
...
 				<div class="title">${translate('layerManager_title')}</div>
-				<ul class="layers">
+				<ul class="layers draggable">
```
thiloSchlemmer, line 183: the draggable attribute in line 189 should be sufficient, for a full drag&drop-funcitonality of the layers, or not?

I've tested with Chrome and FF so far

taulinger (approved):
@alsturm  @thiloSchlemmer: Nicely done!

PR #1988, opened 2023-09-29
ElevationProfile_optimize performance 

norbertK:


taulinger:
Can confirm, works much faster now 🥳

src/modules/elevationProfile/components/panel/ElevationProfile.js
```diff
@@ -331,10 +332,32 @@ export class ElevationProfile extends MvuElement {
...
+							currentColors[selectedAttribute] = {};
+						}
+						if (!currentColors[selectedAttribute].backgroundColor) {
+							if (context.chart.chartArea) {
+								currentColors[selectedAttribute].backgroundColor = this._getBackground(context.chart, elevationData, selectedAttribute);
+							} else {
+								return ElevationProfile.BACKGROUND_COLOR;
+							}
+						}
+						return currentColors[selectedAttribute].backgroundColor;
```
thiloSchlemmer, line 346: Do we need these checks? Can we reduce this to `ElevationProfile.BACKGROUND_COLOR`

thiloSchlemmer (commented):
see inline comments

src/modules/elevationProfile/components/panel/ElevationProfile.js
norbertK: hm - the background part is entered twice for every point (about 6000 times over long distances), but it doesn't 'do' anything, it just returns what is computed in _getBackground (normally ElevationProfile.BACKGROUND_COLOR or _getTextTypeGradient for surface) once.

thiloSchlemmer (approved):
Thx

PR #1999, opened 2023-10-10
fix EWKT export for multiple geometries

### LINKED ISSUE #1993: EWKT export: wrongt result for multiple geometries
It is possible to export a drawing to an EWKT file.
But when I try to import the same exported EWKT file the import fails with the message "the file type is not supported".

Issue:
When dragging the EWKT file (.txt) into the application the following D&D message is shown:
"KML/GPX/GeoJSON/EWKT Datei hierhin ziehen" which communicates to the user that an EWKT import is supported.

taulinger:
fixes #1993

### DIFF
--- a/src/services/ExportVectorDataService.js
+++ b/src/services/ExportVectorDataService.js
@@ -167,10 +167,10 @@ export class OlExportVectorDataService {
 	}
 
 	// todo: refactor to ewkt.js or an ewkt-provider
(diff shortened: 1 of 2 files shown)


thiloSchlemmer (dismissed):
Thx

thiloSchlemmer (approved):
Thx

PR #2000, opened 2023-10-11
adjust beforeUnload behavior

taulinger:
Relevant active tools are now  `[Tools.DRAW, Tools.MEASURE, Tools.MEASURE]` only

### DIFF
--- a/src/plugins/BeforeUnloadPlugin.js
+++ b/src/plugins/BeforeUnloadPlugin.js
@@ -4,6 +4,7 @@
 import { BaPlugin } from './BaPlugin';
 import { $injector } from '../injection';
 import { observe } from '../utils/storeUtils';
+import { Tools } from '../domain/tools';
 
 /**
  * This plugin registers a "beforeunload" event listener when a tool is active and removes it when the no tool is currently active.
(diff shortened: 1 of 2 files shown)


src/plugins/BeforeUnloadPlugin.js
```diff
@@ -36,4 +37,8 @@ export class BeforeUnloadPlugin extends BaPlugin {
 			observe(store, (state) => state.tools.current, onToolChanged, false);
 		}
 	}
+
+	_getTools() {
+		return [Tools.DRAW, Tools.MEASURE, Tools.MEASURE];
```
thiloSchlemmer, line 42: could be reduced to `[Tools.DRAW, Tools.MEASURE]`

thiloSchlemmer (commented):
is there any need to double 'Tools.MEASURE'?

src/plugins/BeforeUnloadPlugin.js
taulinger: Thx must be `ROUTING` of course

thiloSchlemmer (approved):
Thx

PR #2026, opened 2023-10-26
Drawing: Show text under symbol instead of inside symbol

evgeniTUM:
Currently when creating a symbol with text, the text is shown inside the symbol:
Screenshot from 2023-10-26 16-15-58

Show the text under the symbol:
Screenshot from 2023-10-26 16-22-07


taulinger (approved):
Nice, thx

PR #2027, opened 2023-10-27
indicate when a feature is selectable

taulinger:
Changes the cursor when a vector feature or a pixel of a WMS or tile image is selectable.
The default cursor will change to a pointer (👆) in the following cases:

grafik

grafik


thiloSchlemmer (approved):
Thx

PR #2042, opened 2023-11-06
extend MvuContentPanel

taulinger:
Allows lazy loading of the content now:

```
export class MyPanel extends AbstractMvuContentPanel {
	

	createView(model) {
		const { active } = model;
		

		return active ?  ...

```


thiloSchlemmer (approved):
Nice!

PR #2082, opened 2023-11-28
v4.1: fix elevation profile slope zero

taulinger:
* add fix for slope values of zero

* update slope value check


thiloSchlemmer (approved):
Thx

PR #2087, opened 2023-11-29
replace toggle- with switch-component

thiloSchlemmer:


src/modules/utils/components/showCase/ShowCase.js
```diff
@@ -254,13 +254,8 @@ export class ShowCase extends BaElement {
 						<input type="range" />
 					</div>
 
-					<h3>Toggle-Button</h3>
+					<h3>Toggle-Button (Switch)</h3>
```
taulinger, line 257: <h3>Switch</h3>

taulinger (approved):
Nice one!

PR #2108, opened 2023-12-06
manage routing-related query parameters

taulinger:
- [x] update query parameters when route changes
- [x] initially load a route from query parameters
- [x] load a route during the session from query parameters when user activates the rooting tool


thiloSchlemmer (approved):
Nice!

PR #2126, opened 2023-12-14
RoutingChip: add edit behavior

thiloSchlemmer:
image


taulinger:
When a route already exists in the store, clicking on the chip should just activate the routing tool,
```
	setCurrentTool(Tools.ROUTING);
```
 not add a new proposal:



grafik

taulinger (commented):
When a route already exists in the store, clicking on the chip should just activate the routing tool again,
```
	setCurrentTool(Tools.ROUTING);
```
 not add a new proposal:



grafik

taulinger:
Does this component need the exported `RoutingStartStates`  enum`?

taulinger:
When only one waypoint exists (start or destination), the chip is not visible:

grafik

grafik

taulinger (changes requested):
When only one waypoint exists (start or destination), the chip is not visible:

grafik

grafik

taulinger (commented):
Sorry about the confusion, after playing around with the chip I think the best solution would be to restore the behavior from v3: Clicking the chip will always start a new route, and an existing one will be deleted.

```
import { reset, setProposal } from '../../../../store/routing/routing.action';

const Update_Coordinate = 'update_coordinate';
const Update_Status = 'update_status';

/**
 * An AssistChip to start a routing with a proposal {@link module:domain/coordinateTypeDef~Coordinate}
 * @class
 * @extends {AbstractAssistChip}
 * @property {module:domain/coordinateTypeDef~Coordinate} coordinate the coordinate which should be used as a routing point(start or destination)
 * @author thiloSchlemmer
 */
export class RoutingChip extends AbstractAssistChip {
	constructor() {
		super({
			coordinate: []
		});
		const { TranslationService } = $injector.inject('TranslationService');
	}

	update(type, data, model) {
		switch (type) {
			case Update_Coordinate:
				return { ...model, coordinate: [...data] };
		}
	}

	getIcon() {
		return routingSvg;
	}

	getLabel() {
		return translate('chips_assist_chip_start_routing_here');
	}

	isVisible() {
		return true;
	}

	onClick() {
		const { coordinate } = this.getModel();
		reset();
		setProposal(coordinate, CoordinateProposalType.START_OR_DESTINATION);
	}

	static get tag() {
		return 'ba-routing-chip';
	}

	set coordinate(value) {
		this.signal(Update_Coordinate, value);
	}
}
```

src/modules/chips/i18n/chips.provider.js
```diff
@@ -6,6 +6,7 @@ export const provide = (lang) => {
 				chips_assist_chip_elevation_profile: 'Elevation Profile',
 				chips_assist_chip_export: 'Export',
 				chips_assist_chip_start_routing_here: 'Plan a route',
+				chips_assist_chip_edit_existing_route: 'Edit route',
```
taulinger, line 9: Now unneeded

src/modules/chips/i18n/chips.provider.js
```diff
@@ -19,6 +20,7 @@ export const provide = (lang) => {
 				chips_assist_chip_elevation_profile: 'Geländeprofil',
 				chips_assist_chip_export: 'Export',
 				chips_assist_chip_start_routing_here: 'Route planen',
+				chips_assist_chip_edit_existing_route: 'Route bearbeiten',
```
taulinger, line 23: Now unneeded

test/modules/chips/components/RoutingChip.test.js
```diff
@@ -54,64 +53,23 @@ describe('RoutingChip', () => {
...
 	});
 
 	describe('when chip is clicked', () => {
-		it('changes proposal coordinate on click', async () => {
-			const state = { routing: { status: RoutingStatusCodes.Start_Destination_Missing } };
-
+		it('changes to START_OR_DESTINATION proposal coordinate on click', async () => {
 			const properties = { coordinate: coordinate };
-			const element = await setup(state, properties);
+			const element = await setup(properties);
```
taulinger, line 72: Do we have a check that the routing s-o-s was reset?

taulinger (approved):
Finally, thx! 🤩

PR #2132, opened 2023-12-15
adjust routing plugin

taulinger:
when the routing tool is activated, it also

- closes the context menu
- removes existing highlight features


thiloSchlemmer (approved):
Thx

PR #2138, opened 2023-12-19
v4.1: StyleService - add type attribute detection

taulinger:
* add style detection for feature-attribute 'type'

* update text-color for markerstyle

* update marker styles

* update calculation of contrast color

* add/update tests


thiloSchlemmer (approved):
Thx

PR #2156, opened 2024-01-09
RouteInfo:update view while stats are missing

### LINKED ISSUE #2149: Routing: reset UI
when no route could be calculated (due to an error or improper waypoints)

grafik

Should be like this in that case:

grafik

thiloSchlemmer:
resolves #2149 partly


taulinger (approved):
Thx

PR #2157, opened 2024-01-09
Routing: reset stats in case of errror

### LINKED ISSUE #2149: Routing: reset UI

taulinger:
resolves #2149, additionally,  fixes a problem with the `OlElevationProfileHandler`


thiloSchlemmer (approved):
Thx

Issue #2096, opened 2023-12-04
optimize IframeGenerator component


taulinger:
The container's height changes when using the switch

![grafik](https://github.com/ldbv-by/bav4/assets/49945713/92979fd3-6c8a-47b3-8052-bd1ce46a4850)

PR #2175, opened 2024-01-12
Optimize IframeGenerator component

alsturm:
fixes: #2096


taulinger (changes requested):
One test of `Switch.js` is broken now

taulinger (approved):
THX

PR #2198, opened 2024-02-01
configurable response interceptor

taulinger:
Extend the `HttpService` for registering a response interceptor.
A request interceptor may be supported in the future.


thiloSchlemmer (approved):
Nice!

Issue #2189, opened 2024-01-24
Routing and ElevationProfile have different data


alsturm:
![numbers](https://github.com/ldbv-by/bav4/assets/76113751/c46c988f-491e-4ab2-8a03-d10c6ce0f65d)

thiloSchlemmer:
The wording should also be harmonized:

![image](https://github.com/ldbv-by/bav4/assets/2117317/5f520c9f-5734-4d59-b2e9-827949c7a46f)

![image](https://github.com/ldbv-by/bav4/assets/2117317/7d04a5cf-864d-44c7-95e7-d9b259dd200c)

PR #2240, opened 2024-02-28
CoordinateService: add getLength/getArea methods

thiloSchlemmer:
resolves partially #2189 

- [x] coordinateService.getLength()
- [x] coordinateService.getArea() 
- [x] usage in routeStats.provider
- [x] usage in elevation.provider


src/services/OlCoordinateService.js
```diff
@@ -212,4 +214,61 @@ export class OlCoordinateService {
...
+					return sum;
+				}
+				const next = coordinates[index + 1];
+				const r = wgs84.Inverse(current[1], current[0], next[1], next[0]);
+
+				return sum + r.s12;
+			}, 0);
+		};
+		const wgs84Coordinates =
+			coordinateRepresentation?.code && coordinateRepresentation?.code !== 4326
```
taulinger, line 239: `coordinateRepresentation?.code && coordinateRepresentation?.code !== 4326`

- unneeded check for code 4326 (ol.transform can safely always be called even if the coordinates are already in the target SRID)
- I would suggest throwing an error when `CoordinateRepresentation` contains no `srid` property

thiloSchlemmer: ok, so....

- coordinateRepresentation is not optional
- and something like `if(!coordinateRepresentation.code) throw new Error('The coordinates must have a reprojectable CoordinateRepresentation (code!== null)')`

taulinger (commented):
grafik

taulinger (commented):
grafik

taulinger (approved):
Yeah, thx

PR #2259, opened 2024-03-13
MapService: add calcArea() and calcLength()

taulinger:
This PR gives the `MapService` two high-level methods for length and area calculation based on  an array of coordinates.
The MapService detects which kind of calculation is needed and does all the necessary transformation of the coordinates.
Then it calls the low-level methods of the `OlCoordinateService`. For `getLength` its implementation could be:

```
/**
	 * Calculates the length of an array of coordinates.
	 *
	 * The kind of calculations depends on the  `geodesic` parameter:
	 * When `geodesic` is set to `true`, the coordinates must be either in 4326 or 3857 and calculation is done in a geodesic manner.
	 * When `geodesic` is set to `false`, the coordinates must be transformed in a projected SRID before and calculation is done on a projected plane.
	 *
	 *
	 * @param {Array<module:domain/coordinateTypeDef~Coordinate>} coordinates input coordinates
	 * @param {boolean} [geodesic=true] `true` if calculation should be done in geodesic manner
	 * @returns {Number} the length
	 */
	getLength(coordinates, geodesic = true) {
		const getGeodesicLength = (coordinates) => {
			const wgs84 = Geodesic.WGS84;
			return coordinates.reduce((sum, current, index, coordinates) => {
				if (index === coordinates.length - 1) {
					return sum;
				}
				const next = coordinates[index + 1];
				const r = wgs84.Inverse(current[1], current[0], next[1], next[0]);

				return sum + r.s12;
			}, 0);
		};
		return geodesic ? getGeodesicLength(coordinates) : new LineString(coordinates).getLength();
	}
```

### DIFF
--- a/src/services/MapService.js
+++ b/src/services/MapService.js
@@ -206,4 +206,62 @@ export class MapService {
 			left: visibleRectangle.left - baseRectangle.left
 		};
 	}
+
+	/**
+	 * Calculates the length of an array of coordinates.
+	 *
+	 * The kind of calculations depends on whether the coordinates are inside or outside of the extent
+	 * of the supported local projected system (if defined).
+	 *
+	 * This is basically a convenience method for {@link module:services/OlCoordinateService#getLength}.
+	 *
+	 * @param {Array<module:domain/coordinateTypeDef~Coordinate>} coordinatesInMapProjection coordinates in map projection
+	 * @returns {Number} the length
+	 */
+	calcLength(coordinatesInMapProjection) {
+		if (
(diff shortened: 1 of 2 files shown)


thiloSchlemmer (approved):
Looks good to me. This makes olGeometryUtils.getProjectedLength/getProjectedArea obsolete (in Branch [dev_CoordinateService_getLength-getArea](https://github.com/ldbv-by/bav4/tree/dev_CoordinateService_getLength-getArea), which isn't the worst thing either

thiloSchlemmer (approved):
Nice!

PR #2271, opened 2024-03-19
Badge text 

alsturm:


src/modules/toolbox/components/toolBar/ToolBar.js
taulinger, line 1: Uses i18n key `header_logo_badge_standalone` from the `header` module which is not allowed
So let's extend the toolbox i18n for this entry...

taulinger (approved):
THX

PR #2276, opened 2024-03-21
ElevationProfile: improve performance

### LINKED ISSUE #2270: ElevationProfile: slow performance on large distance (from routing)
When creating elevation profiles from routing geometries,  the tooltip and highlight-features seems to lag or get frozen.

taulinger:
We avoid pushing huge arrays of coordinates to the store. Detecting changes can be expensive in that case. Instead, we reference them by an identifier and push this identifier to the store.

Fixes #2270


thiloSchlemmer (approved):
Nicely done! 

Cannot find a lag anymore. One single long running task belongs to the chart.js component and  occurs only while th init-phase of the chart.

PR #2287, opened 2024-03-25
GeoResourceResultItem: support GeoResource roles

taulinger:
- [x] extend GeoResourceResultItem
- [x] after sign-out remove layers referencing non-allowed GeoResources


thiloSchlemmer (approved):
Nice

PR #2290, opened 2024-03-27
remove non-allowed GeoResources on sign-out only when initiated by the user

taulinger:
Session invalidation due to session timeout should keep the active GeoResources, so the user can continue without re-selecting them


thiloSchlemmer (approved):
Thx

PR #2299, opened 2024-04-04
WaypointItem: remove display of coordinate

thiloSchlemmer:
avoid the [horror vacui](https://en.wikipedia.org/wiki/Horror_vacui) and remove the unusable, spacefilling display of waypoint coordinates


taulinger (approved):
Thx

PR #2333, opened 2024-04-17
ShareService: remove all html paths (not only index.html)

### LINKED ISSUE #1973: IFrame Generator: wrong 'iframe' src link created when run locally
How to reproduce:
1. Share
2. Create IFrame Preview

Result: The generated code contains an invalid link:
`<iframe src=http://localhost:8081/embed.html/embed.html?...`

It seems that this happens only when run locally. If deployed the problem does not occur.

taulinger:
fixes #1973


thiloSchlemmer (approved):
Thx

PR #2334, opened 2024-04-18
DrawTool: configure tools by parameter

thiloSchlemmer:


src/modules/iframe/components/tools/DrawTool.js
```diff
@@ -61,6 +64,18 @@ export class DrawTool extends MvuElement {
...
+			// The list of toolNames should contain at least one defined toolName, otherwise the tools will not be modified
+			const valid = (candidates) => tools.some((tool) => candidates.includes(tool.name));
+
+			return valid(toolNames)
+				? tools.map((tool) => {
+						return { ...tool, visible: toolNames.includes(tool.name) };
+					})
+				: tools;
+		};
+
```
taulinger, line 78: Can we be case-insensitive for the given value here?
 ```html
<bayern-atlas
ec_draw_tool="polygon"
ec_draw_tool="POLYGON"
ec_draw_tool="Polygon"
></bayern-atlas>
```

File: test/modules/iframe/components/DrawTool.test.js [hunk 2]
```diff
@@ -74,6 +74,107 @@ describe('DrawTool', () => {
...
 		});
 
+		describe('QueryParameters.EC_DRAW_TOOL is a comma-separated list of values', () => {
+			it('builds the full list of tools (from invalid toolName)', async () => {
+				const queryParam = new URLSearchParams(`${QueryParameters.EC_DRAW_TOOL}=fooBar`);
+				spyOn(environmentServiceMock, 'getQueryParams').and.returnValue(queryParam);
+				const element = await setup();
+
+				expect(element.shadowRoot.querySelectorAll('.draw-tool__button')).toHaveSize(3);
+				expect(element.shadowRoot.querySelector('#point-button')).toBeTruthy();
```
taulinger, line 84: Just a hint: For a long time we check the presence of an element with
```
expect(element.shadowRoot.querySelectorAll('#point-button')).toHaveSize(1)
```
cause when doing so we also ensure that there exist no doublets of this element

taulinger (approved):

PR #2340, opened 2024-04-22
IFrame:  use dndImportPanel

thiloSchlemmer:


taulinger (approved):
Thx, will merge it also now myself, needed for our current WC styling dev

PR #2343, opened 2024-04-23
MeasurementOverlayStyle: handle pointerup event on map once while dragging overlay

thiloSchlemmer:
fix handling of dragPan interaction while dragging an overlay


taulinger (approved):
THX

PR #2347, opened 2024-04-23
public WC: rework style setup

taulinger:
grafik


src/modules/public/components/publicComponent.css
alsturm, line 1: maybe add a default value:

	height: 100%;
	width: 100%;

alsturm (approved):

alsturm (approved):

PR #2375, opened 2024-05-08
olGeometryUtils: update behavior of multiLineStringToLineString

### LINKED ISSUE #2368: olGeometryUtils may return invalid coordinates
Requesting a profile for https://www.geodaten.bayern.de/freizeitwege/KML/114.kml leads to invalid coordinates.

thiloSchlemmer:
fixes #2368

Updates the behavior of multiLineStringToLineString to return null, when the given geometry did not match the geometry validation.

Valid geometry is:
- GeometryType -> MultiLineString
- all parts must be connected by its start and end nodes


taulinger (approved):
THX

PR #2381, opened 2024-05-15
ElevationProfile: fix gradient calculation 

thiloSchlemmer:
Fix the calculation of gradiant colorstops to maintain correct x-coordinate relative to the elements distance from profile start.


taulinger (approved):
THX

PR #2440, opened 2024-06-14
RouteChart: optimize display of route segment type 'unknown' [Ticket#45104573,#45104574]

thiloSchlemmer:
- optimize display of route segment type 'unknown'
- fix formatting of distance values in chart tooltip


taulinger (approved):
THX

PR #2458, opened 2024-06-24
FeatureInfoItem: remove simplify of elevation profile coordinates [Ticket#45104576]

thiloSchlemmer:
Remove the process step of simplifying coordinates for the elevation profile, due to the fact that the profile provider does this implicit now. Any simplification beforehand is unnecessary.


taulinger (approved):
THX

PR #2471, opened 2024-07-02
GlobalErrorPlugin: remove all associated layers from the layers s-o-s

taulinger:
When a Georesource is unloadable, all referenced layers should be removed to keep the UI in sync


thiloSchlemmer (approved):
Thx!

PR #2484, opened 2024-07-09
adding mapListener after old features are read

taulinger:
* adding mapListener after old features are read

* update OlMEasurementHandler and OlDrawHandler, to prevent multiple unnecessary calls of _save/store

* Ensure that transferred features are saved


thiloSchlemmer (approved):
Thx

PR #2508, opened 2024-07-18
increase WMS timeout and finetune error handling

taulinger:
* BvvBaaImageLoadFunction: increase timeout

* remove layer only when GeoResource requires authentication


thiloSchlemmer (approved):
Thx

PR #2511, opened 2024-07-19
always preseve the crosshair marker when sharing the map

taulinger:
This PR extends the CROSSHAIR query param by its coordinate to preserve the crosshair when shared after the map was relocated
```
http://localhost:8080/?crh=true,1340544.8149054,6318369.4588622
```
The coordinate is always in 3857.


thiloSchlemmer (dismissed):
Thx

thiloSchlemmer (approved):
Thx

PR #2565, opened 2024-09-10
v4.2: ContextMenu - display parcel number when available

taulinger:
display parcel number when available


thiloSchlemmer (approved):
Thx!

PR #2573, opened 2024-09-20
do not open the FeatureInfo tab when no result is available

taulinger:
Avoid opening the FeatureInfo tab when it is closed and no feature info was found

grafik


thiloSchlemmer (approved):
Thx

PR #2606, opened 2024-10-07
rework the bottomSheet slice-of-state

taulinger:
It's not easy in tests to detect when a certain bottom sheet is displayed and when not.
Therefore the following changes are made:

1. The `active` property is an array now and reflects all active bottom sheets. An empty array means no bottom sheet is active. This is the most important change
2.  A `BottomSheetContent` item will be removed from the `data` property when `closeBottomSheet` is called


thiloSchlemmer (approved):
Thx!

thiloSchlemmer (approved):
Nice!

PR #2598, opened 2024-10-04
OlMfpHandler: update mfp-layer render behavior

thiloSchlemmer:
- add getRenderPixel() function to fix the calculation of the mfp mask
- rearrange the render methods

fixes #2595

PR #2609, opened 2024-10-08
4.2 OlMfpHandler: update mfp-layer render behavior (#2598)

thiloSchlemmer:
* fix mpf-layer render behavior

- add getRenderPixel() function to fix the calculation of the mfp mask
- rearrange the render methods


taulinger (approved):
THX

PR #2622, opened 2024-10-10
TimeTravelPlugin:  Set the timeTravel active property to false when die BottomSheet could not be opened

taulinger:
Set the timeTravel active property to false when die BottomSheet could not be opened


thiloSchlemmer (approved):
THX

PR #2636, opened 2024-10-17
Add internal chip

alsturm:


taulinger (commented):
There seems to be a broken test

taulinger (approved):
THX

PR #2689, opened 2024-11-08
RoutingPlugin: restore  lost feature (#107939)

taulinger:
When we have waypoints for a route but the tool should not be active, we deactivate the tool after the route is successfully fetched


thiloSchlemmer (approved):
Thx

PR #2695, opened 2024-11-12
ElevationProfile:fix chartjs configuration for x axe

thiloSchlemmer:
do not display the given max value, although the chart must be streched to this value


taulinger (approved):
THX

PR #2715, opened 2024-11-21
Mfp3Encoder:  fix shortenedUrl for qrCode

thiloSchlemmer:
reset QueryParameter 'TOOL_ID' to get a url without active ExportTool


taulinger (approved):
THX

PR #2761, opened 2024-12-10
HelpTooltip: change positioning from center-left to top-left 

thiloSchlemmer:
for better readability of distance overlay
**Before**
center-left
**Optimized**
top-left


taulinger (approved):
Nice, should we add this vo v4.3 also?

PR #2773, opened 2024-12-13
Iframe accessibility

alsturm:


thiloSchlemmer:
Looks good to me, at this point.  We should think about or discuss [ARIA role types](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles#aria_role_types) for our custom components, which are accessible through the embeded mode (`ba-icon` -> role: button or img? )

taulinger (approved):
Looks formally good to me!

PR #2787, opened 2024-12-18
StyleService:  handle text feature with empty text style

thiloSchlemmer:
* use default-style if text style is missing

* use default-style if image style is missing


taulinger (approved):
THX

PR #2808, opened 2025-01-10
Filestorage: adjust debounce delay and timeout

taulinger:
* increase debounce time from 1000 to 2000ms

* increase POST  timeout from 1000 to 2000ms


thiloSchlemmer (approved):
THX

PR #2812, opened 2025-01-10
MeasurementOverlayStyle: prevent overlays minimum lenght

thiloSchlemmer:
prevent display of measurement (partition) overlays while geometry length is smaller then 1000 meters


taulinger (approved):
THX

PR #2818, opened 2025-01-14
DrawToolContent: expand property form after 'modify'

thiloSchlemmer:
add behavior to expand feature property form after first change to 'modify' (#2811)


taulinger (approved):
THX

PR #2820, opened 2025-01-15
encode URL of imported vector data again

taulinger:
* fix forUrl(): take the URL as ID again if not specified

* improve doc


thiloSchlemmer (approved):
Thx

PR #2822, opened 2025-01-16
kml: prevent 'null' for description-tag (4.3)

thiloSchlemmer:
update cleaning rules for kml format to prevent `null` values in name and description tag


taulinger (approved):
THX

PR #2824, opened 2025-01-17
kml: prevent 'null' for description-tag

thiloSchlemmer:
update cleaning rules for kml format to prevent null values in name and description tag (#2822)


taulinger (approved):
thx

PR #2827, opened 2025-01-21
DrawTool: display geometry statistic 

thiloSchlemmer:
add statistic to draw-s-o-s and display in draw tool content


src/modules/olMap/handler/draw/OlDrawHandler.js
```diff
@@ -525,7 +544,7 @@ export class OlDrawHandler extends OlLayerHandler {
 			this._draw.abortDrawing();
 			this._modify.setActive(false);
 			setSelection([]);
-
+			setStatistic(defaultDrawStats);
```
taulinger, line 547: Can't see a test for mutating the draw s-o-s

src/modules/toolbox/components/drawToolContent/DrawToolContent.js
taulinger, line 1: No test needed for its changes?

taulinger (commented):
Hmm, do we really need to extend also the measurement s-o-s for the `statistic` property?

src/domain/geometryStatisticTypeDef.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * @module domain/geometryStatisticTypeDef
+ */
+/**
+ * Statistic data of a geometry.
+ * @typedef {Object} GeometryStatistic
```
taulinger, line 6: Are all properties mandatory? If not so, let's mark them as optional.

src/domain/geometryStatisticTypeDef.js
```diff
@@ -0,0 +1,11 @@
...
+ * @property {module:domain/coordinateTypeDef~Coordinate} coordinate the coordinate of the feature
```
taulinger, line 7: What does the coordinate mean when the geometry is a line or polygon?

taulinger (approved):
THX

PR #2859, opened 2025-01-28
FileStoragePlugin: also set  the fileID

taulinger:
* remove early save after reading rocksi features again

* change and rename action `setAdminId` to `setAdminAndFileId`

* set both the admin and the file ID if the admin id is available


thiloSchlemmer (approved):
thx

PR #2878, opened 2025-02-05
OlMeasurementHandler: optimize behavior after measurement is activated

thiloSchlemmer:
When measuring starts and old features are read in, which are only drawings, the handler activates the draw-interaction. If at least one measurement feature exists in the old features, the handler activates the modify-interaction.


taulinger (approved):
THX

PR #2884, opened 2025-02-06
HttpService: increase default timeout

taulinger:
* increase default timeout from 1s to 10s

* remove timeouts where appropriate  and < 10s

* make prettier happy


thiloSchlemmer (approved):
Thx

PR #2903, opened 2025-02-13
CoordinateInfo: display as compact component

thiloSchlemmer:


src/modules/info/components/coordinateInfo/CoordinateInfo.js
```diff
@@ -29,7 +31,9 @@ export class CoordinateInfo extends MvuElement {
 	constructor() {
 		super({
 			coordinate: null,
-			elevation: null
+			displaySingleRow: false,
```
taulinger, line 34: `displaySingleRow` could also be a private field of this component

src/modules/info/components/coordinateInfo/CoordinateInfo.js
taulinger, line 1: Class has a poor test coverage

taulinger (approved):
THX

PR #2953, opened 2025-03-10
Use raster layer as default bg layer again

taulinger:
Its VectorTile pendant is only used if we have a retina display
This reverts commit 099d0998bd198f2b1f7996862a075a84eff7bfc7.


thiloSchlemmer (approved):
Thx

PR #2955, opened 2025-03-11
OlDrawHandler: fix behavior on pointer click

thiloSchlemmer:
If the user starts a drawing and clicks on a measurement geometry, the draw process should not interrupt, but proceed with changed draw state.


taulinger (approved):
THX

PR #2963, opened 2025-03-13
Units: update bvv distance formatting (4.3)

thiloSchlemmer:
distance < 1km: 1 fraction digit
distance > 1km: 2 fraction digits


taulinger (approved):
THX

PR #2977, opened 2025-03-18
 VectorGeoResource: add support for BaFeature

taulinger:
- [x] `VectorGeoResource` can have `BaFeature`s as data source
- [x] rename `Feature` to `BaFeature`; `Geometry` to `BaGeometry`
- [x] introduce a generic `StyleHint` which can be applied to a `VectorGeoresource`, `RtVectorGeoresource` or `BaFeature`
- [x] extend the `StyleService` for that purpose
- [x] improve jsdoc for all `GeoResources`

### DIFF
--- a/src/domain/feature.js
+++ b/src/domain/feature.js
@@ -3,7 +3,7 @@
  */
 
 import { isString } from '../utils/checks';
-import { Geometry } from './geometry';
+import { BaGeometry } from './geometry';
 
 /**
  * A feature.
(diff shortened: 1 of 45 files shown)


src/domain/feature.js
```diff
@@ -3,22 +3,23 @@
  */
 
 import { isString } from '../utils/checks';
-import { Geometry } from './geometry';
+import { BaGeometry } from './geometry';
 
 /**
  * A feature.
```
thiloSchlemmer, line 9: Suggestion: 'A feature in BA context' - to clarify that there are also other 'feature' (OpenLayers).

thiloSchlemmer (changes requested):
Looks good to me. The class jsdoc for `BaFeature` needs more info, if possible.

thiloSchlemmer (approved):
Nice

PR #2986, opened 2025-03-25
handle authorization downgrade

taulinger:
We show the Sign-In dialog and do a retry of the intercepted requests


thiloSchlemmer (approved):
Thx

PR #3023, opened 2025-04-11
SearchResultsPanel: finetune behavior while switching between mouse and keyboard

thiloSchlemmer:


taulinger (changes requested):
Nice, but does not work as expected in some cases

taulinger (approved):
Nice 😍, thx

PR #3035, opened 2025-04-15
MeasurementOverlayStyle: fix behavior on resolution:change

thiloSchlemmer:
fix behavior on resolution:change, adopt old properties while update style


taulinger (approved):
THX

PR #3087, opened 2025-05-13
MeasurementOverlayStyle: handle empty overlays property correctly

thiloSchlemmer:

PR #3088, opened 2025-05-13
MeasurementOverlayStyle: handle empty overlays property correctly

thiloSchlemmer:
cherry pick from (#3087)

handle empty overlays property correctly


taulinger (approved):
THX

PR #3089, opened 2025-05-13
StyleService: add OlStyleService

### LINKED ISSUE #3081: Refactor VectorLayerService and StyleService
Goal: Remove the tight coupling of both services

Approach:

1. Add a new StyleService2 with only four public methods
2. Move all the logic from calling `_applyStyle` and so on to  the StyleService

Image

thiloSchlemmer:
resolves #3081


taulinger (commented):
Well done!

PR #3092, opened 2025-05-16
Fix of Common Component Searchable Select

herrmutig:
- Fixed: event target was not same as the element that invoked the event - led to unresponsive clicking behaviour.
- Fixed: dropdown crops if had not enough space from its parent element.


taulinger (approved):
THX

taulinger (approved):

PR #3107, opened 2025-05-27
Mfp3Encoder: add 'OAF' GeoResourceType as exportable

thiloSchlemmer:


taulinger:
grafik

Hmm, let's try to display the same name shown in the LayerItem ("Baudenkmäler" in this case)

```js
geoResourceService.byId(olLayer.get('geoResourceId')).label
```

taulinger (commented):
Works for me (see inline comments), but the message seems to be outdated ("The following layers.."):

grafik

PR #3108, opened 2025-05-27
Search keybord navigation optimization

alsturm:


taulinger (commented):
Works like a charm for me!

1. In Chrome the highlight seems to be lost in bright mode

image

2.  Maybe we should check for the `shiftKey` property to prevent the navigation when the text should be just marked (@thiloSchlemmer but maybe that's another PR just for v4.5)

image

taulinger (approved):
Well done!

PR #3127, opened 2025-06-11
allow reset of OAF filter

taulinger:


test/store/layers/layers.reducer.test.js
```diff
@@ -793,6 +793,10 @@ describe('layersReducer', () => {
 			modifyLayer('id0', { filter: 'cql' });
 
 			expect(store.getState().layers.active[0].constraints.filter).toBe('cql');
+
+			modifyLayer('id0', { filter: null });
```
herrmutig, line 797: test with { filter: '' } for modifyLayer is missing 

Rest looks good to me!

herrmutig (approved):

PR #3136, opened 2025-06-16
encode/decode layer filter from query parameter

taulinger:


src/utils/urlUtils.js
```diff
@@ -78,3 +78,21 @@ export const setQueryParams = (url, params = {}) => {
...
+ * Note:
+ * It provides an alternative for `URLSearchParams`.
+ *
+ * While URLSearchParams is suitable for decoding URL queries, for encoding it can lead to unexpected results such as spaces being encoded as `+` and extra characters such as `~` being percent-encoded (it uses `application/x-www-form-urlencoded`).
+ * Therefore, this method encodes each entry of the the given query parameters by calling `encodeURIComponent`.
+ *
+ * @param {object} queryParameters The query parameters for that query string
+ * @returns {string} the query parameters string
+ */
+export const queryParamsToString = (queryParameters) => {
```
herrmutig, line 94: jsdoc "@function" is missing for const lamdas in the file. (s. e.g oafUtils.js)

src/plugins/LayersPlugin.js
```diff
@@ -73,6 +82,9 @@ export class LayersPlugin extends BaPlugin {
 								if (!!layerStyle[index] && isHexColor(layerStyle[index])) {
 									atomicallyAddedLayer.style = { baseColor: layerStyle[index] };
 								}
+								if (!!layerFilter[index] && isString(layerFilter[index]) && layerFilter[index].length) {
```
herrmutig, line 85: `!!layerFilter[index]` probably not needed. `isString(layerFilter[index])` ensures existence and type.

PR #3155, opened 2025-06-25
LayerItem: improve handling of settings sheet

taulinger:
- [x] close active settings sheet when sheet of another layer item is opened
- [x] avoid re-opening an already open settings sheet


thiloSchlemmer (approved):
Thx

PR #3160, opened 2025-06-27
ShareApi: use fallback dialog (v4.4)

thiloSchlemmer:
refactor share-api usage to use fallback dialog instead of warning notification


taulinger (approved):
THX

PR #3179, opened 2025-07-02
LayerItem: display layerStates

thiloSchlemmer:
update behavior and ui of LayerItem while layerState is changing


taulinger (commented):
Nice!

PR #3206, opened 2025-07-14
optimize OAF loading

taulinger:
- handle optional OAF response properties `numberReturned` and `numberMatched`
- limit OAF request via debounce


thiloSchlemmer (approved):
Thx

PR #3219, opened 2025-07-15
LayerSettingsPanel

thiloSchlemmer:
Adding a new component to change  layer settings like:

- color (layer/georesource basecolor)
- updateinterval


taulinger:
The interval of the GeoResource does not seem to be properly adopted 

grafik

src/modules/layerManager/components/LayerSettingsPanel.js
```diff
@@ -0,0 +1,237 @@
...
+import { DEFAULT_MIN_LAYER_UPDATE_INTERVAL_SECONDS } from '../../../domain/layer';
+
+const Update_Layer_Settings = 'update_layer_Settings_State';
+
+/**
+ * State of a layer setting
+ * @readonly
+ * @enum {String}
+ */
+export const SettingState = Object.freeze({
```
taulinger, line 19: Do we need this enum to be exported?

test/modules/layerManager/components/LayerSettingsPanel.test.js
```diff
@@ -0,0 +1,230 @@
...
+			const element = await setup(layer);
+
+			//view does not contain any element
+			expect(element.shadowRoot.querySelectorAll('.layer_setting').length).toBe(0);
+			expect(element.shadowRoot.querySelectorAll('.layer_setting_title').length).toBe(0);
+			expect(element.shadowRoot.querySelectorAll('.layer_setting_content').length).toBe(0);
+			expect(element.shadowRoot.querySelectorAll('ba-switch').length).toBe(0);
+		});
+
+		it('does not renders the view with invalid layerId (no GeoResource)', async () => {
```
taulinger, line 130: typo: .> does not render[]

taulinger (commented):
Overall, it looks good, but we have some issues left (see comments)

taulinger:
Should we display the values from the GeoResource in those cases?
That means the UI is displayed as "active", but `modifyLayer` is only called when the user has actively set a different value

grafik

taulinger (commented):
Very nice, but still a fundamental consideration:
Should we display the values derived from the GeoResource in those cases?
That means the UI is displayed as "active", but `modifyLayer` is only called when the user has actively set a different value

grafik

taulinger (approved):
Well done!

PR #3220, opened 2025-07-15
OAF UI: Zoom to extent feature + description for title attribute

herrmutig:


taulinger (commented):
Looks good!
Additionally, we could give the `<option>` of the `<select>` also the description as title attribute:
Image

taulinger (approved):
Well done!

PR #3222, opened 2025-07-16
fixed: layerId did not refresh UI

### LINKED ISSUE #3221: OafMask does not re-render when `layerId` is set via setter

herrmutig:
resolve #3221


taulinger (approved):
THX

PR #3226, opened 2025-07-17
Export features: remove internal properties before export

### LINKED ISSUE #2930: Export features without unnecessary properties
Features which are created with any tool of the BAV4 contains arbitary properties, which are only useful in BAV4 context.

This properties should be removed before export.

thiloSchlemmer:
resolves #2930


taulinger:
Don't forget GeoJSON

thiloSchlemmer:
> Don't forget GeoJSON

As long as KML and the function `kml~create()` be the central part of our persisting and storing strategy we should be fine.
Never the less i added a test, which fails if we change the process of saving measurement- or draw-data.

PR #3239, opened 2025-07-23
Fix OafMask overrides filter of wrong layer

### LINKED ISSUE #3231: OafMask gets wrong filter assigned when switching to another (OAF) Layer
How to reproduce:

1. Add two OAF geodata to Layers
2.  Open the OafMask on one of the added layers.
3. Add a filter
4. Open the other Layer

You should now see that the (wrong) filter is applied to the now open OafMask.

herrmutig:
resolve #3231


taulinger (approved):
THX

PR #3256, opened 2025-07-31
layers.reducer: determine resulting style

taulinger:
…ntegrate into reducer


taulinger:
@thiloSchlemmer Next Step: Refactor the `OlStyleService`. The service just needs to handle the steps marked red

grafik

PR #3294, opened 2025-09-19
AdminCatalog - Visual Hint and Loading Spinner

herrmutig:


src/modules/admin/components/AdminCatalog.js
```diff
@@ -111,14 +120,32 @@ export class AdminCatalog extends MvuElement {
...
+		const droppedBranch = this.#lastDroppedBranch;
+		this.#lastDroppedBranch = null;
+
+		if (droppedBranch) {
+			const domBranch = this.shadowRoot.querySelector(`#catalog-tree-root li[branch-id="${droppedBranch.id}"] .catalog-branch`);
+			domBranch.classList.add('branch-added');
+
+			setTimeout(() => {
+				domBranch.classList.remove('branch-added');
+			}, 1000);
```
taulinger, line 140: Let's check if this part is really needed

src/modules/admin/components/adminCatalog.css
```diff
@@ -87,6 +108,10 @@ select {
 	padding: 0 15px 0 0;
 }
 
+.branch-added {
+	animation: branch-added-animation 1s;
```
taulinger, line 112: Would suggest `5s`

PR #3357, opened 2025-10-16
Fix styling of feature or layer

taulinger:
This PR fixes two issues regarding the style of a Feature or VectorLayer

- highlighting a feature does not affect its original style
- a default style is not applied to a layer when its GeoResource is not styleable


thiloSchlemmer (approved):
Thx!

PR #3362, opened 2025-10-17
HTMLPrintService and FeatureInfo printing

herrmutig:
- Service that allows printing specific HTML content with the browser's default pdf printer
- Print Feature for FeatureInfos

### DIFF
--- a/src/injection/config.js
+++ b/src/injection/config.js
@@ -98,7 +98,7 @@ $injector
 	.register('ExportVectorDataService', OlExportVectorDataService)
 	.register('ImportWmsService', ImportWmsService)
 	.registerSingleton('ImportOafService', new ImportOafService())
-	.register('HtmlPrintService', new HtmlPrintService())
+	.register('HtmlPrintService', HtmlPrintService)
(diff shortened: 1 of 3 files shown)


src/main.css
```diff
@@ -222,3 +222,12 @@ body,
 	-webkit-user-select: none;
 	user-select: none;
 }
+
+/* 
+ Used for pdf printing. Prints only elements inside #html-print
```
taulinger: `Used for PDF printing directly from a component, e.g., FeatureInfoPanel. Prints only elements inside #html-print`

src/services/HtmlPrintService.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * @module services/HtmlPrintService
+ */
+
+import { render } from 'lit-html';
+
+/**
+ * @class
```
taulinger: Maybe we should add a brief description of what this service does

taulinger (commented):
A very smart solution!

taulinger (approved):

PR #3363, opened 2025-10-17
WebComponent

taulinger:


src/modules/wc/components/PublicWebComponent.js
```diff
@@ -0,0 +1,583 @@
...
+			case QueryParameters.EC_SRID: {
+				const validSRIDs = [4326, this.#mapService.getSrid(), this.#mapService.getLocalProjectedSrid()].map((n) => n.toString());
+				return this.#passOrFail(() => validSRIDs.includes(attr.value), `Attribute "${attr.name}" must be one of [${validSRIDs.join(',')}]`);
+			}
+			case QueryParameters.EC_GEOMETRY_FORMAT: {
+				const validFormats = [SourceTypeName.EWKT, SourceTypeName.GEOJSON, SourceTypeName.KML, SourceTypeName.GPX];
+				return this.#passOrFail(() => validFormats.includes(attr.value), `Attribute "${attr.name}" must be one of [${validFormats.join(',')}]`);
+			}
+		}
+		return false;
```
thiloSchlemmer: Using the DEFAULT case, opens the opportunity to be expressive (with a additional comment) in a convienient way of the structure of the switch clause too:

```suggestion
        default:
            // we ignore all other attribute candidates
            return false;
			}
		}
		
```

src/services/EnvironmentService.js
```diff
@@ -7,6 +7,7 @@ import { $injector } from '../injection';
  * @class
  */
 export class EnvironmentService {
+	#embedRegex = /(\/embed[/]?(index.html)?|embed.html)$/;
```
thiloSchlemmer: Not used anymore

PR #3383, opened 2025-10-30
provide GeoResource  info from GeoResource itsself

taulinger:


src/domain/geoResources.js
```diff
@@ -215,6 +216,14 @@ export class GeoResource {
 		return this._attributionProvider;
 	}
 
+	/**
+	 * Returns a description n about this GeoResource
+	 *  @type {String|null}
```
herrmutig: let's keep it like the other "string" types lower-cased. `@type{string|null}`

PR #3396, opened 2025-11-10
ToolContent: add collaborative editing badge and related functionality to Draw- & MeasureTool

thiloSchlemmer:
image


taulinger (commented):
Looks and works nicely!
We could move `people.svg` (used in three locations) to the global asset folder, right?

PR #3428, opened 2025-12-01
AdminCatalog - Validation and confirmation for tree groups

herrmutig:
Changes:
- To delete a group the user must now confirm the action.
- Group name validation (must contain at least 1 letter to allow confirmation)


taulinger (commented):
Nice, thx

PR #3442, opened 2025-12-12
extend WebComponent examples

taulinger:
- [x] add example using the drawing tool
- [x]  allow adding a previously created KML
- [x] fire baGeometryEvents in a debounced manner


thiloSchlemmer (approved):
Nice

Issue #3459, opened 2026-01-08
Web Component Optimizations


herrmutig:
- [ ] Trim and parse the "l" attribute into an array (ignore spaces around commas). E.g. `l="GEORESOURCE_AERIAL,GEORESOURCE_WEB,GEORESOURCE_HISTORIC"` => `l="GEORESOURCE_AERIAL, GEORESOURCE_WEB, GEORESOURCE_HISTORIC"`
- [ ] Evaluate if BA events can be registered outside of `ba-load` and document expected behavior.
- [ ] Add an example/guide to replicate a current state of  BayernAtlas Frontend (URI-Parameters).
- [ ] Update API docs and improve wording to state that attributes are included in `event.detail` at events.
- [ ] Add a short description for domain specific terms (Feature, Layer, GeoResource, Geometry).
- [ ] Document `baChange` event properties to answer what properties will invoke a `baChange` event

PR #3465, opened 2026-01-12
Fixes for PublicWebComponent

### LINKED ISSUE #3458: Web Component - Bug List
- [x] `modifyLayer(id, {visible:false|true})`  does not fire a `baChange` event
- [x]  `modifyLayer(id, {opactiy:0.5})`  does not fire a `baChange` event
- [x]  Properties `layersVisiblity` and `layersOpacity` throw a NullReferenceException when accessed in baLoad,  expected to return an `[]:bool` filled with initial layers
- [x]  If no layers exist anymore the `ba.layers` property should return an empty array. -> Currently: `['']`, Expected: `[]`

taulinger:
fixes #3458
see also #3459

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
 				"dompurify": "3.3.1",
 				"geographiclib-geodesic": "2.2.0",
 				"lit-html": "3.3.2",
-				"maplibre-gl": "5.15.0",
+				"maplibre-gl": "5.16.0",
 				"ol": "^10.7.0",
 				"proj4": "2.20.2",
 				"redux": "5.0.1",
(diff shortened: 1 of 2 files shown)


src/modules/wc/components/PublicWebComponent.js
```diff
@@ -348,55 +367,72 @@ export class PublicWebComponent extends MvuElement {
...
+		);
 	}
 
 	/**
-	 * Returns the opacity of the layers of the map
+	 * Returns the opacity of the layers of the map or `[]` if the map is not yet initialized.
+	 *
 	 * @type {Array<number>}
 	 */
 	get layersOpacity() {
```
herrmutig: If WC is initialized like this: `<bayern-atlas l=""></bayern-atlas>`  then 
`layersOpacity`  returns an array with one entry:`[NaN]` . Expected `[]`

PR #3476, opened 2026-01-19
Baselayerswitcher setup

alsturm:


src/modules/baseLayer/components/switcher/assets/14.webp
taulinger: Why are the image pixel sizes (width x height) so different?

Shouldn't they all be the same dimensions?

PR #3518, opened 2026-02-04
Copy to clipboard functionality for BuildInfo (Showcase) 

herrmutig:
Adds a clipboard Icon to the build-info button to easily copy the current build information to the clipboard.

ctc-bi


taulinger (commented):
Nicely done!
Maybe we should also add a tooltip ("Open the Showcase") now for opening the showcase dialog:

Image

PR #3729, opened 2026-04-10
HighlightPlugin: enhance highlight feature management

taulinger:
After re-opening the FeatureInfo panel, the previously existing highlight should be restored

grafik


herrmutig (approved):
Looks Good, thanks!

PR #3758, opened 2026-04-22
v4.5: retrieve name and description from GeoJSON and GPX sources 

taulinger:
VectorLayerService: get  MetaData for GeoJSON and GPX sources


thiloSchlemmer (approved):
Thx

PR #3767, opened 2026-04-29
ClusterStyle: use contrastColor consequent

thiloSchlemmer:
Using contrastColor not only for the number, but the border.

image


taulinger (approved):
Thx!

PR #3846, opened 2026-06-19
display the route even without elevation information

taulinger:
Previously, a route was only displayed if the corresponding elevation data (profile) could be determined. This is not always possible outside of Bavaria. Now, the route is displayed regardless, even though certain statistical values ​​may be omitted.

grafik

image


thiloSchlemmer (approved):
Thx!

PR #3847, opened 2026-06-22
Add dependency-cruiser for dependency analysis and update scripts

taulinger:
- Added "dependency-cruiser" version 17.4.3 to package.json
- Introduced a new script "doc:deps" to generate a dependency graph in SVG format


herrmutig (approved):
Looks good, Thanks!

PR #3857, opened 2026-06-25
olGeometryUtils: add upper boundary for getSimplifiedLineString (v4.5)

thiloSchlemmer:
* add upper boundary for getSimplifiedLineString


taulinger (approved):
THX

PR #3860, opened 2026-06-29
OlStyleService: cache & reuse anchor values

thiloSchlemmer:
To prevent losing information about anchor-values of an OpenLayers IconStyle, we save them as internal property.

#45120762


taulinger (approved):
THX

PR #3868, opened 2026-07-01
KML: add internal property normalized anchor

thiloSchlemmer:


taulinger (changes requested):
As discussed yesterday...

src/utils/propertyUtils.js
```diff
@@ -44,7 +44,7 @@ export const isInternalProperty = (propertyKey) => {
 };
 
 /**
- * Defines a list of internally used legacy feature property keys that should not be displayed in an UI or exported
+ * Defines a list of internally used (legacy) feature property keys that should not be displayed in an UI or exported.
```
taulinger: The changed comment should also be reverted

taulinger (approved):
THX

PR #3889, opened 2026-07-08
olStyleUtils: Fix  measurement style function with map rotation

### LINKED ISSUE #3872: Graphical error during measurement
Measure using a ruler while aligned to the north. Then rotate—moving out of the north alignment. That is when the small ruler markings/unit ticks begin to shift.

Image

thiloSchlemmer:
fixes #3872


taulinger (approved):
Nice!

PR #3908, opened 2026-07-21
Open Legend Button on Home Tab

herrmutig:


taulinger (commented):
Well done!
We should just make the legend icon a bit smaller so that it matches the size of the other icons.

Image

Image

taulinger (commented):
Image

Image

taulinger (approved):
THX

PR #3915, opened 2026-07-23
Elevation profile with line of sight

thiloSchlemmer:
image
image
image


taulinger (commented):
Wow, great!

Two things caught my attention:

1. The tab should remain active when changing the measurement line.

2. I don't quite understand the green dot. The target isn't visible in the screenshot. Nevertheless, it's displayed.

image

thiloSchlemmer:
> Wow, great!
> 
> Two things caught my attention:
> 
> 1. The tab should remain active when changing the measurement line.
> 2. I don't quite understand the green dot. The target isn't visible in the screenshot. Nevertheless, it's displayed.
> 
> image

1. You're right. It took a while before it started to annoy me. I run out of time and thought, this could be the next optimization :)

2. The green circle (with the check mark) appears, if the target is visible to the oberver. 
Other cases are:
-  the target is not visible, but the distance under the line of sight is in a reasonable limit (currently 1000m) to think of  a building or tower which could be build to show up in the line of side
- the target is not visible and the line of sight goes into the space. 

The (other) green circle is just the observer, which can have a more visual effect, if the observer height is used for buildings/towers or whatever (90m, 195m) to be separated from the rendered line of sight.  

Examples: 
- Line of sight from north tower of Frauenkirche (Munich) to Aufhofen -> Observer should stand at 95m above elevation.
- Line of sight from Munich Olympiaturm (195m) to Zugspitze

src/services/ElevationService.js
```diff
@@ -12,6 +12,7 @@ import { indicateChange } from '@src/store/elevationProfile/elevationProfile.act
 /**
  * @typedef {Object} Profile
  * @property {Array<ProfileElevation>} elevations elevations objects of this profile
+ * @property {Array<module:domain/coordinateTypeDef~CoordinateLike>} sourceCoordinates sourceCoordinates on which the elevations are based
```
taulinger:
```js
@property {Array<module:domain/coordinateTypeDef~Coordinate>} sourceCoordinates 
```

taulinger (commented):
That’s going to be great.

taulinger (approved):
Pop it in.

PR #3924, opened 2026-07-27
MFP-Export support for Legends

herrmutig:


taulinger (approved):
Works perfectly. Very nice!

PR #4003, opened 2026-09-16
olDraw: update selection style correctly

### LINKED ISSUE #3761: Draw Tool - Dots do not change color immediately
Image
Image

How to reproduce:
- Select draw tool (line or polygon tool)
- Draw a shape
- Deselect and Select the shape, thus making the dots visible
- In the tool settings, change the style to another color. 

The lines and fill color will update, but the dots will stay the same color as shown in the image.

thiloSchlemmer:
fixes #3761


src/modules/olMap/handler/measure/OlMeasurementHandler.js
```diff
@@ -627,7 +627,7 @@ export class OlMeasurementHandler extends OlLayerHandler {
 		select.getFeatures().on('remove', (e) => {
 			const feature = e.element;
 			const styles = feature.getStyle();
-			feature.setStyle(styles.slice(0, -SELECT_STYLES_COUNT));
+			feature.setStyle(styles.filter((style) => !(style instanceof NamedStyle && ['Selection', 'ConstructionLine'].includes(style.name))));
```
taulinger, line 630: Is no adjustment to the test required here?

taulinger (approved):

